PR #43, opened 2025-02-23
Add wanaku router web UI

ibek:


orpiske:
This looks great! As it's still WIP, I'm marking as draft, just to prevent us from merging it by mistake.

Issue #76, opened 2025-02-27
[UI] Add quarkus-quinoa extension


lordrip:
Install the quarkus quinoa extension to build the UI when building the application.

The extension should be installed here: https://github.com/wanaku-ai/wanaku/tree/main/routers/wanaku-router

[Quinoa docs](https://quarkus.io/extensions/io.quarkiverse.quinoa/quarkus-quinoa/)

orpiske:
🎉 This issue has been resolved in `v0.0.1` ([Release Notes](https://github.com/wanaku-ai/wanaku/releases/tag/v0.0.1))

PR #92, opened 2025-02-28
feat(UI): Add quinoa

lordrip:
### Context
Currently, the UI is built separately from the project.

This commit adds the quarkus quinoa extension to build the UI as part of the project build.

### How to use
Launch the project in dev mode, and as part of the `wanaku-router` model, the UI will be started in `dev` mode.
When building for production, the UI will be built and added to the final artifact.

fix: https://github.com/wanaku-ai/wanaku/issues/76


orpiske (approved):
LGTM, thanks. Ideally the quinoa version should be put in the parent pom, so it's easier to manage, but I can fix this later. 

I'll give it a test and then merge, so we can have this on the release

ibek (approved):

PR #100, opened 2025-03-03
Automatically generate stub API for UI

Croway:
As per quarkus docs I added the openapi docs into `src/main/webui/` and for the time being I think this is ok to have the generated spec in the repository, in the future, we may refactor quinoa and oval to use the target folder.

I am not a UI expert.. I am open to comments (best practices, naming convention..).

The usage of the generated stub is the following, for example, tools list
```
  getApiV1ManagementTargetsToolsList().then((response) => {
    // Update the model with the response
    ...
  });
  ```

### CODE REVIEW

File: ui/yarn.lock [hunk 1]
```diff
@@ -1202,6 +1568,246 @@ __metadata:
...
+    "@stoplight/spectral-ref-resolver": "npm:^1.0.4"
+    "@stoplight/spectral-runtime": "npm:^1.1.2"
+    "@stoplight/types": "npm:~13.6.0"
+    "@types/es-aggregate-error": "npm:^1.0.2"
+    "@types/json-schema": "npm:^7.0.11"
+    ajv: "npm:^8.17.1"
+    ajv-errors: "npm:~3.0.0"
+    ajv-formats: "npm:~2.1.0"
+    es-aggregate-error: "npm:^1.0.7"
+    jsonpath-plus: "npm:10.2.0"
```
orpiske, line 1551: This one is bringing a CVE, so the PR is failing due to that

orpiske (changes requested):
Apart from the CVE, it looks OK to me, but I need a second look from @lordrip or @ibek who understanding the UI bits lot more than I do.

orpiske:
Thanks for your contribution!

orpiske (approved):
LGTM

ibek (approved):
+1 to merge. I believe we should use hooks instead of calling the API directly. This way, updates only require changes in the hooks, keeping the components clean. We can implement this in ui in upcoming PRs

orpiske:
Thanks, let's merge!

PR #110, opened 2025-03-04
Add Resources List

Croway:
This is an attempt to use the generated API stub, and this is what I noticed:

* https://github.com/wanaku-ai/wanaku/compare/ui-resources-list?expand=1#diff-45a778cbdc19b5343e80bee3e46f743425d943658b941b9e827865f763718434R64 - when using Response as return type, we should always document the actual return type like this
* I am not sure this is ok to have it here https://github.com/wanaku-ai/wanaku/compare/ui-resources-list?expand=1#diff-3550a8c8813bc37aef982dcb8f241c1756429b859b5ffc0ab9190cc2e3fdb522R27 
* I created a .env file to externalize the backend endpoint with the following content `VITE_WANAKU_ROUTER_BASE_URL=http://localhost:8080` it is ignored via gitignore, is it expected?  Moreover, I wonder why VITE expects env to start with VITE, I wanted to name it `WANAKU_ROUTER_BASE_URL` but it seems like vite prevents it OOB, I am probably missing something here.

@lordrip @ibek any suggestion is welcome, moreover the client generation we are using support fetch as well, wdyt? should we switch to fetch instead of axiom?

This is a screenshot of the page with an actual API call

image


servers/wanaku-server-quarkus/src/main/java/ai/wanaku/server/quarkus/api/v1/resources/ResourcesResource.java
```diff
@@ -58,6 +61,8 @@ public Response expose(ResourceReference resource) {
     @Path("/list")
     @GET
     @Produces(MediaType.APPLICATION_JSON)
+    @APIResponse(responseCode = "200", content = @Content(
+            schema = @Schema(type = SchemaType.ARRAY, implementation = ResourceReference.class)))
```
lordrip: I think using `RestResponse`, automatically generates the return type.

```java
import org.jboss.resteasy.reactive.RestResponse;

public RestResponse<InvoiceEntity> get(@PathParam("companyId") String companyId, @PathParam("id") Longid) { }
```

services/providers/wanaku-provider-s3/src/main/java/ai/wanaku/provider/s3/S3ResourceDelegate.java
```diff
@@ -7,6 +7,7 @@
 import ai.wanaku.core.services.provider.AbstractResourceDelegate;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
```
lordrip: Is this import needed?

ui/src/App.tsx
```diff
@@ -21,8 +21,11 @@ import {
 import { Link, Outlet } from "react-router-dom";
 import "./App.scss";
 import { ExternalLinks, Links } from "./router/links.models";
+import axios from "axios";
 
 function App() {
+  axios.defaults.baseURL = import.meta.env.VITE_WANAKU_ROUTER_BASE_URL || 'http://localhost:8080';
```
lordrip: I think we could use the env variable directly `VITE_API_URL` because it's populated by vite itself here: https://github.com/wanaku-ai/wanaku/blob/ce17e4749829df0aedcd24c23e9f81f037e1dadb/ui/vite.config.ts#L12

When building it with quarkus, quinoa sets the information here: https://github.com/wanaku-ai/wanaku/blob/ce17e4749829df0aedcd24c23e9f81f037e1dadb/routers/wanaku-router/src/main/resources/application.properties#L37 so it's relative to the project URL.

services/providers/wanaku-provider-s3/src/main/java/ai/wanaku/provider/s3/S3ResourceDelegate.java
Croway: ops, no

servers/wanaku-server-quarkus/src/main/java/ai/wanaku/server/quarkus/api/v1/resources/ResourcesResource.java
Croway: RestResponse can help, but the API should be changed, in particular, right now we are returning a list in case of 200ok, and a string in case of error, this is not possible with `RestResponse<List<ResourceReference>>`, I cannot return `RestResponse.ResponseBuilder.serverError().entity("Failed to list resources").build()` in case of exception

ui/src/App.tsx
Croway: I am using VITE_API_URL, but it still seems a bit hacky, the `replaceAll` in particular

ibek:
+1 for fetch. I tried to use the generated apis via hooks in #113  and I had to actually update this generated part to make types working .. maybe also something to configure the client generation: https://github.com/wanaku-ai/wanaku/pull/113/files#diff-fcb7564ecb73fa697d55adbf9b5974cc47309f8acd113886fcf403799e5c4a31

servers/wanaku-server-quarkus/src/main/java/ai/wanaku/server/quarkus/api/v1/resources/ResourcesResource.java
orpiske: My .2 cents: better change the API now to make it adequate than later when we have to worry with breaking things for a larger group of users. So I am OK with this, as far as I am concerned.

orpiske (approved):

servers/wanaku-server-quarkus/src/main/java/ai/wanaku/server/quarkus/api/v1/resources/ResourcesResource.java
ibek: +1, I faced the same problem with ToolsList, missing the info in the generated client API

ibek: so I did the same for tools here: https://github.com/wanaku-ai/wanaku/pull/113/files#diff-c649126c7ff75bbfd15343a50f30fe0cf7b9fffac107a7eaa5874d3d35029e05

ui/src/App.tsx
```diff
@@ -24,7 +24,7 @@ import { ExternalLinks, Links } from "./router/links.models";
 import axios from "axios";
 
 function App() {
-  axios.defaults.baseURL = import.meta.env.VITE_WANAKU_ROUTER_BASE_URL || 'http://localhost:8080';
+  axios.defaults.baseURL = import.meta.env.VITE_API_URL.replaceAll("\"", "");
```
lordrip: I think we don't need the `import.meta.env` part because we're defining the variable using the definePlugin from vite.

```suggestion
  axios.defaults.baseURL = VITE_API_URL.replaceAll("\"", "");
```

Croway: interesting, I'll try that. Thanks

Croway: nice, VITE_API_URL did the trick, and I could remove the replaceAll(..)

ibek (approved):

orpiske:
🎉 This issue has been resolved in `v0.0.2` ([Release Notes](https://github.com/wanaku-ai/wanaku/releases/tag/v0.0.2))

PR #113, opened 2025-03-04
UI resources and tools hooks for api calls

ibek:


ui/src/Pages/Tools/ToolsPage.tsx
```diff
@@ -67,46 +44,19 @@ export const ToolsPage: FunctionComponent = () => {
...
-            description: "The count of facts to retrieve",
-          },
-        },
-        required: ["count"],
-      },
-    },
-  ];
 
-  const data = fetchedData || testData;
+  const data = fetchedData || [];
```
lordrip: [Suggestion]: Considering when `fetchedData` is `null`, a new empty array is created, what if we set `[]` as seed value for the `useState` directly?  This way we don't need this `data` variable and we won't be creating new arrays.

```tsx
const [fetchedData, setFetchedData] = useState<ToolReference[]>([]);

// Then use `fetchedData` instead of `data`
```

lordrip: There are a couple of things we might want to consider about this component.

1. We're optionally assigning `ToolsList` depending on whether there's data or not, perhaps a better approach might be extracting the `ToolList` JSX part into a separate component and receive `fetchedData` as a prop, this way the `ToolsPage` component is not only smaller but easier to reason about because we removed the conditional assignment.
2. We have 2 modals defined in this component and we're reading the form value by hand using the browser API, in this case, this could be counterproductive because we're mixing React scheduling with DOM scheduling, meaning that we're assuming the code will run exactly in order, which might not be the case, especially for modest clients. An alternative would be to extract the internals of each `Modal` into separate components and use state and callbacks to each of them, this way we delegate entirely to React when the data is available and updated, removing the need to query the DOM directly.

That being said, I understand this PR is not changing nor touching those behaviors (they were already there prior to this PR) so please take my comments with a :pinching_hand: of salt and use it for food for thought :smile:

lordrip (approved):
I left a couple of comments, please feel free to take them into account or to delay them to another PR :pray:

servers/wanaku-server-quarkus/src/main/java/ai/wanaku/server/quarkus/api/v1/tools/ToolsResource.java
```diff
@@ -58,6 +63,7 @@ public Response add(ToolReference resource) {
     @Path("/list")
     @GET
     @Produces(MediaType.APPLICATION_JSON)
+    @APIResponse(responseCode = "200", content = @Content(schema = @Schema(type = SchemaType.ARRAY, implementation = ToolReference.class)))
```
Croway: In order to be consistent with the resource code, we should add 2 APIResponse, one for 200, and one for 500, ala

```
@Path("/list")
    @GET
    @Produces(MediaType.APPLICATION_JSON)
    @APIResponses({
            @APIResponse(responseCode = "200", content = @Content(
                    schema = @Schema(type = SchemaType.ARRAY, implementation = ResourceReference.class))),
            @APIResponse(responseCode = "500", content = @Content(schema = @Schema(implementation = String.class)))
    })
```
I'll update https://github.com/wanaku-ai/wanaku/pull/110#issuecomment-2698566585

ibek: Thanks, I added it there

Croway (approved):

orpiske:
🎉 This issue has been resolved in `v0.0.2` ([Release Notes](https://github.com/wanaku-ai/wanaku/releases/tag/v0.0.2))

PR #118, opened 2025-03-05
Use fetch instead of axios

Croway:
The changes on the APIResponse side (handle of 200 and 500) introduce this behavior https://github.com/wanaku-ai/wanaku/compare/use-fetch?expand=1#diff-6d938346c493d6e4fccb956db2669b49a691d287f684f4858af5fe6ae8bd8856R15 I do not like it that much, do you think it can be improved somehow?


Croway:
> > The changes on the APIResponse side (handle of 200 and 500) introduce this behavior https://github.com/wanaku-ai/wanaku/compare/use-fetch?expand=1#diff-6d938346c493d6e4fccb956db2669b49a691d287f684f4858af5fe6ae8bd8856R15 I do not like it that much, do you think it can be improved somehow?
> 
> @Croway, interesting, why is that? 🤔

Due to https://github.com/wanaku-ai/wanaku/blob/main/servers/wanaku-server-quarkus/src/main/java/ai/wanaku/server/quarkus/api/v1/resources/ResourcesResource.java#L48 as you can see, right now the API can reply in 2 ways, with a `ResourceReference[]` in case of successful execution, or a `String` in case of internal server error. This is ported to the openapi spec, and the code generation handle `ResourceReference[] | string`

Croway:
With the latest commit I casted the object  to the expected one in the pages, but I think we need to discuss error/exception handling, and how to represent them in the UI.

For example, I introduced a custom fetch client, would it make sense to centralize there error handling?

ibek:
Suggested by AI but how about to have one common response type like this? 

Wrap your response in a consistent envelope so that both success and error responses share the same base type. For example, create a response class:

java
public class ResourceListResponse {
  private ResourceReference[] data;
  private String error; // null on success
  // getters/setters
}
Then change your API annotations to use that type for both responses:

java
@APIResponses({
  @APIResponse(responseCode = "200", content = @Content(
      schema = @Schema(implementation = ResourceListResponse.class))),
  @APIResponse(responseCode = "500", content = @Content(
      schema = @Schema(implementation = ResourceListResponse.class)))
})
public Response list() {
  // return a ResourceListResponse with data populated on success or error set on failure
}
Now the generated code will consistently return a ResourceListResponse and you can do:

ts
setFetchedData(result.data);

lordrip:
@ibek I think it makes sense, an alternative with quarkus is to have:

```java
import org.jboss.resteasy.reactive.RestResponse;

public RestResponse<MenuEntity>
```

whereas MenuEntity could be whatever response type we want, the advantage that I see vs having an annotation is that the method is type safe, and also it's automatically populated to the OpenAPI schema at once, if I'm not mistaken, the annotation could be made with a type, and the method could be something completely different,

That being said, both scenarios will need to deal with the error handling on the UI, in some backends, there's no HTTP errors per see, but rather responses with errors in it, both method (the one you brought and this one) "forces" the UI to be HTTP error-aware, which is not a bad thing, just something to consider when dealing with backend responses.

Croway:
Hi @ibek @lordrip We could do the following:

* Get rid of the APIResponses annotation by returning a RestResponse<T>.
* Get rid of catching exceptions and return a custom message, to do this we should always use RuntimeExceptions, and let an ExceptionMapper create the error response.

let's discuss the T now, we have two options:
* An Object that includes the data and the error (as suggested by @ibek), this way the openapi contains all the objects, for example `record WanakuResponse<T>(WanakuError error, T data)`, where WanakuError is a simple POJO (for now) `record WanakuError(String message)`. This is verbose, the list signature is `public RestResponse<WanakuResponse<List<ResourceReference>>> list()` :) 
* Always return what's included in the successful response, in our case, `RestResponse<List<ResourceReference>>` this way the openapi spec does not contain all the object, and we'll need to handle this in the UI.

Croway:
The lates [commit](https://github.com/wanaku-ai/wanaku/pull/118/commits/ac4a2f4582ff390e38761a11cae0ef1105b41e16) contains the first option

orpiske:
Just ack'ng that I'll take a closer look at this one today. I was busy on the f2f

servers/wanaku-server-quarkus/src/main/java/ai/wanaku/server/quarkus/api/v1/resources/ResourcesResource.java
```diff
@@ -45,18 +47,13 @@ public Response expose(ResourceReference resource) {
...
-            @APIResponse(responseCode = "200", content = @Content(
-                    schema = @Schema(type = SchemaType.ARRAY, implementation = ResourceReference.class))),
-            @APIResponse(responseCode = "500", content = @Content(schema = @Schema(implementation = String.class)))
-    })
-    public Response list() {
+    public RestResponse<WanakuResponse<List<ResourceReference>>> list() {
         try {
             List<ResourceReference> list = resourcesBean.list();
-            return Response.ok(list).build();
+            return RestResponse.ok(new WanakuResponse<>(list));
```
ibek: @Croway will you do the same for Tools list api?

Croway: sure, I'll think a bit more about the available options, and then I'll update all the pages.

orpiske (approved):
LGTM, thanks!

Croway:
@ibek @orpiske @lordrip 

I've updated Tools and did a minor refactor to the API:

- API Responses with Object (not 204), now should use WanakuResponse<T>
- API code should handle only the 2xx path, in case of error, just throw the exception, and let the ExceptionMappers take care of that
-  Created ExceptionMappers, a general one for all the Exceptions except the Wanaku ones that simply reply with error 500 and `generic error` text (mostly for security reasons, we do not want the client to recieve stacktraces messages :)), and a Wanaku one that replies with the actual exception message and error 500, for more contextual information.
-  Aligned Tools, Resources, Target APIs, UI, and CLI

@ibek I hope I didn't break the tool part (the table works correctly for me), if you can do a quick test that would be nice

PR #123, opened 2025-03-07
Simplifies service management

### LINKED ISSUE #78: Simplify management of targets
* We should have a centralized registry of targets 
* Services should register on startup and deregister on shutdown 
* Mapping should be done via dns and services referred via their URIs (ie; `wanaku:file` or something similar)

orpiske:
- Move the registry of services to an external Valkey instance
- Register the instances on service startup
- Remove the link/unlink APIs and commands

This resolves #78.


orpiske:
@ibek @lordrip @Croway With this one we don't have to run `wanaku targets tools link` nor `wanaku targets resources link` anymore. The downstream services register automatically in the Valkey (open source version of Redis). 

It's a large one, so likely will conflict with #118

Issue #90, opened 2025-02-28
Automate release with jreleaser and Maven


orpiske:
We need to investigate why this JReleaser configuration is not working with Maven. 

```xml
<profile>
      <id>release</id>
      <build>
          <plugins>
              <plugin>
                  <groupId>org.jreleaser</groupId>
                  <artifactId>jreleaser-maven-plugin</artifactId>
                  <version>${jreleaser-maven-plugin.version}</version>
                  <inherited>false</inherited>
                  <configuration>
                      <gitRootSearch>true</gitRootSearch>
                      <jreleaser>
                          <release>
                              <github>
                                  <owner>wanaku-ai</owner>
                                  <overwrite>true</overwrite>
                                  <changelog>
                                      <formatted>ALWAYS</formatted>
                                      <preset>conventional-commits</preset>
                                  </changelog>
                              </github>
                          </release>
                          <distributions>
                              <app>
                                  <type>BINARY</type>
                                  <brew>
                                      <active>ALWAYS</active>
                                  </brew>
                                  <artifacts>
                                      <artifact>
                                          <path>cli/${distribution.directory}/{{distributionName}}-{{projectVersion}}-linux-x86_64.tar.gz</path>
                                          <platform>linux-x86_64</platform>
                                      </artifact>
                                      <artifact>
                                          <path>${distribution.directory}/{{distributionName}}-{{projectVersion}}-windows-x86_64.zip</path>
                                          <platform>windows-x86_64</platform>
                                      </artifact>
                                      <artifact>
                                          <path>${distribution.directory}/{{distributionName}}-{{projectVersion}}-osx-x86_64.zip</path>
                                          <platform>osx-x86_64</platform>
                                      </artifact>
                                  </artifacts>
                              </app>
                          </distributions>
                      </jreleaser>
                  </configuration>
              </plugin>
          </plugins>
      </build>
  </profile>
```

It fails with: 

```
[ERROR] No plugin found for prefix 'jreleaser' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/Users/opiske/.m2/repository), central (https://repo.maven.apache.org/maven2)] 
```

There are [some](https://github.com/jreleaser/jreleaser/discussions/1237) notes on the web about specific changes to apply for multi-module projects with JReleaser and it seems we are missing something.

orpiske:
[...]

PR #172, opened 2025-03-15
issue: Automate release with jreleaser and Maven #90

SherineL00187650:
Made following changes to the POM.xml file to resolve the issue

1.  JReleaser plugin is added in the parent "POM.xml" under the <pluginManagement> section.
2.  Added "jreleaser-maven-plugin.version" property is set corrected "pom.xml"


orpiske:
Thanks for your contribution! I will take a look at it during the week

orpiske:
Thank you! I will investigate this in the next few days.

PR #178, opened 2025-03-17
File Persistence and MongoDB persistence

Croway:
Persistence core module, at the moment 2 persistence flavour are implemented:

- File, the default one, is enabled if there are no other persistence-related beans in the context. This flavor works like the actual persistence on file system
- MongoDB, uses quarkus-mongodb-client, and is enabled if the property `wanaku.persistence.mongodb` is set to `true`


ibek (approved):

orpiske (changes requested):
Overall, it looks good to me and I think you are on the right track. 

There's a few improvements I think we could have on the code:
* try to add javadoc to the core persistence api 
* I'd prefer if we named it `core-persistence-file`, `core-persistence-api` and `core-persistence-mongodb`: the reason being that as we add more and more components for the code, it starts to get a bit confusing to navigate in the project tree. 
*  Make sure to add the MongoDB container (or Postgres) in the `docker-compose.yml` file. 
* Also make sure to add documentation about this on the usage guide.

Thanks!

core/core-service-discovery/src/main/java/ai/wanaku/core/service/discovery/ValkeyRegistry.java
```diff
@@ -1,249 +0,0 @@
...
- * wanaku-target-address -> ip:port
- * wanaku-target-type -> tool-invoker
- *
- * And also the configurations for the service as a hashmap
- * config1 -> description1, config2 -> description2, etc.
- *
- * This class is basically iterating over the hashmap
- */
-@ApplicationScoped
-public class ValkeyRegistry implements ServiceRegistry {
```
orpiske, line 42: I think this change is out of scope for this

core/core-service-discovery/src/main/java/ai/wanaku/core/service/discovery/WanakuServiceRegistry.java
```diff
@@ -0,0 +1,118 @@
...
+            for (Map.Entry<String, String> entry : configurations.entrySet()) {
+                Configuration configuration = new Configuration();
+                configuration.setDescription(entry.getValue());
+                configurationMap.put(entry.getKey(), configuration);
+            }
+            Configurations config = new Configurations();
+            config.setConfigurations(configurationMap);
+            service.setConfigurations(config);
+
+            serviceRepository.persist(service);
```
orpiske: For now, I'd like to keep the services non-persistent. Previously (i.e.; up to 0.0.2) we used to persist then, but now we have auto-registration (using Valkey as the source of state) ... so let's keep them out of persistence for now.

routers/wanaku-router/src/main/resources/application.properties
```diff
@@ -33,7 +33,7 @@ quarkus.grpc.server.use-separate-server=false
 %test.wanaku.router.indexes-path=target/test-data
 
 # Quinoa configuration
-quarkus.quinoa=true
+quarkus.quinoa=false
```
orpiske: Out of scope.

servers/wanaku-server-quarkus/src/main/java/ai/wanaku/server/quarkus/api/v1/management/targets/TargetsBean.java
```diff
@@ -30,20 +32,43 @@ public class TargetsBean {
...
+    @Inject
+    ServiceRepository serviceRepository;
+
     public void configureTools(String service, String option, String value)
             throws IOException {
-        Map<String, String> configurations = toolsConfigurations(service);
-        configurations.put(option, value);
+        // TODO why is it needed?
+//        Map<String, String> configurations = toolsConfigurations(service);
+//        configurations.put(option, value);
```
orpiske: This is loading the service configurations directly from the downstream services and then persisting them. 

I am not sure if we need to keep all of it. We must, however, persist any configuration that had been previously set. For instance, for the Kafka downstream, if the user had set the `bootstrapHost` configuration, then we want to keep that persisted.

servers/wanaku-server-quarkus/src/main/java/ai/wanaku/server/quarkus/api/v1/management/targets/TargetsBean.java
```diff
@@ -30,20 +32,43 @@ public class TargetsBean {
...
+        serviceRepository.update(storedService);
     }
 
     public void configureResources(String service, String option, String value)
             throws IOException {
-        Map<String, String> configurations = resourcesConfigurations(service);
-        configurations.put(option, value);
+        // TODO why is it needed?
+//        Map<String, String> configurations = resourcesConfigurations(service);
+//        configurations.put(option, value);
```
orpiske: It is the same thing as above, but applied for the resources configurations.

services/tools/wanaku-routing-yaml-route-service/src/main/resources/application.properties
```diff
@@ -23,4 +23,7 @@ wanaku.service.routing.base-uri=direct:start
 # valkey.port=6379
 
 # Timeout
-# valkey.timeout=10
\ No newline at end of file
+# valkey.timeout=10
+
+#quarkus.mongodb.connection-string = mongodb://localhost:27017
+#wanaku.persistence.mongodb = true
```
orpiske: I'm a bit confused here. The downstream services should not have access to the DB. The only medium that they use is the service registry (Valkey).

orpiske (changes requested):
Nice to see all the improvements @Croway! There are still some things to tweak, but it's getting better and better. I'm looking forward seeing this one on the code 🙏

servers/wanaku-server-quarkus/src/main/resources/application.properties
```diff
@@ -7,4 +7,7 @@ quarkus.log.category."ai.wanaku".level=DEBUG
 quarkus.log.category."io.quarkiverse.mcp".level=DEBUG
 
 %dev.quarkus.http.cors=true
-%dev.quarkus.http.cors.origins=http://localhost:5173
\ No newline at end of file
+%dev.quarkus.http.cors.origins=http://localhost:5173
+
+#quarkus.mongodb.connection-string = mongodb://localhost:27017
+#wanaku.persistence.mongodb = true
```
orpiske: I assume it defaults to file if not persistence provider is defined, right?

servers/wanaku-server-quarkus/src/test/java/org/wanaku/server/quarkus/support/TestIndexHelper.java
```diff
@@ -18,7 +19,7 @@ public static File createToolsIndex() throws IOException {
 
         indexFile.deleteOnExit();
 
-        if (indexFile.exists()) {
+        if (indexFile.exists() && !Files.readString(indexFile.toPath()).equals("[]")) {
```
orpiske, line 22: What's exactly the check `!Files.readString(indexFile.toPath()).equals("[]")` testing? Is it verifying if it's using MongoDB? If yes, maybe it would be good to move this to a separate method (`isMongoDB`) so it's easier to understand.

services/providers/wanaku-provider-file/src/main/resources/application.properties
```diff
@@ -30,4 +30,7 @@ wanaku.service.provider.service.defaults.idempotent=false
 # valkey.port=6379
 
 # Timeout
-# valkey.timeout=10
\ No newline at end of file
+# valkey.timeout=10
+
+#quarkus.mongodb.connection-string = mongodb://localhost:27017
+#wanaku.persistence.mongodb = true
```
orpiske: I guess this is out of scope, as downstream services should not interact w/ the persistence layer.

services/tools/wanaku-routing-yaml-route-service/src/main/resources/application.properties
Croway: This is not the case anymore, I'll remove these, only the wanaku-server-quarkus itself need the persistence layer, the services only need K/V store

servers/wanaku-server-quarkus/src/main/resources/application.properties
Croway: exactly, I'll add the file defaults as well

servers/wanaku-server-quarkus/src/test/java/org/wanaku/server/quarkus/support/TestIndexHelper.java
Croway: integration tests are executed only with Files at the moment, I modified how the file is created via https://github.com/wanaku-ai/wanaku/pull/178/files#diff-fe8e9904e7d28bb1adf5f14547249fcdcb078d8edfa314504e7dbcf6598c1c24R26 and the tests need to be aligned, but I'll refactor this part

services/providers/wanaku-provider-file/src/main/resources/application.properties
Croway: correct

orpiske:
Per discussion, this should also resolve https://github.com/wanaku-ai/wanaku/issues/190.

orpiske (approved):
This is ready to go, thanks!

orpiske:
🎉 This issue has been resolved in `v0.0.4` ([Release Notes](https://github.com/wanaku-ai/wanaku/releases/tag/v0.0.4))

Issue #155, opened 2025-03-12
Optional parameters must be filtered out


orpiske:
Currently optional parameters are not filtered out if they are not present. This makes many tools harder to define and harder to use.

Consider this URI:

https://swapi.dev/api/people/{id}/?search=Luke+Skywalker

It should become the following if `id` is not available:

https://swapi.dev/api/people/?search=Luke+Skywalker

orpiske:
Obs.: this is an important issue and we need to have it soon. I assigned to me so I don't forget, but it will take a few days until I get back to this one (so, in case anyone wants to take a shot at it, drop me a note and feel free to take it over).

orpiske:
Fixed with the linked PR.

orpiske:
🎉 This issue has been resolved in `v0.0.4` ([Release Notes](https://github.com/wanaku-ai/wanaku/releases/tag/v0.0.4))

PR #187, opened 2025-03-21
RFC: proposed fix for URL parsing

orpiske:
This should fix issue #155.

Folks, I'd like to hear your thoughts about this fix ... 

This patch changes how URI are evaluated. Previously, the code was building the URIs semi-manually. With this patch, it changes to the logic to use [Qute](https://quarkus.io/guides/qute-reference#strict_rendering), from Quarkus. 

This makes the whole rendering of the URI, a lot more powerful ... but, also, a bit more complex.

For instance, the URI in the dog facts API would need to change from: 

```https://dogapi.dog/api/v2/facts?limit={count}```

to:

```https://dogapi.dog/api/v2/facts?limit={count or 1}```

However, this code also makes it possible to build the query part dynamically. For instance, consider a tool with an URI defined as `http://my-host/{query.build}`,  and two arguments named as `id` and `name`.  If receiving the values "456" for the id and the `My Name With Spaces` for the name ... this would result in a URI built dynamically, such as: `http://my-host/?id=456&name=My+Name+With+Spaces` (note also that the values are encoded). The `{query.build}` is the trick that makes it build it dynamically. 

Take a look at the `URIParserTest` for additional examples.


Croway:
LGTM, but I was wondering if there are changes required in https://github.com/wanaku-ai/wanaku-toolsets

luigidemasi (approved):

orpiske:
Thanks for the reviews and comments. I'll merge this once I prepare the PR for the toolsets.

Issue #185, opened 2025-03-21
Add the ability to override global Service configuration per invocation using Tool definition


luigidemasi:
support the configuration override like this one:

```json
    {
        "name": "get_weather", 
        "description": "Get current weather information for a location",
        "inputSchema": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "City name or zip code"
            }
            "_configuration.xxxxx=yyyyy": {
              "type": "string",
              "description": "configuration xxxx for the service"
            }
          },
          "required": ["location"]
        }
      }

```

orpiske:
@luigidemasi I have been thinking about this one and also #186 and #187. One thing that I've considered for #187 was to add `class` and `scope` properties to the arguments. Something like this:


```
{
        "name": "get_weather", 
        "description": "Get current weather information for a location",
        "inputSchema": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "City name or zip code"
              "class": "argument"
              "scope": "llm"
            }
          },
          "required": ["location"]
        }
      }
```

Then, we could bundle all these changes so that 


```
 xxxxx=valueXXXX": {
        "type": "string",
         "description": "configuration xxxx for the service"
         "class": "configuration"
         "scope": "service"
},
 zzzz=valueZZZZ": {
        "type": "string",
         "description": "header zzzz for the service"
         "class": "header"
         "scope": "service"
},
 zzzz=valueZZZZ": {
        "type": "string",
         "description": "header zzzz for the service"
         "class": "configuration"
         "scope": "service-endpoint"
}
```

The idea here is that the `class` identifies the purpose* (i.e.: where to be used) and the `scope` identifies to where it applies (i.e.: `llm` is something to be used by the model, `service` is something to be used by a downstream service`, `server` would be used by Wanaku MCP, and so on). 

What do you think?

With that, we could simplify and consolidate a lot of the handling of the

luigidemasi:
I like it, but I would go further, I would add the `value` property to the body, like:

```json

 "xxxxx": {
        "type": "string",
         "description": "configuration xxxx for the service",
         "class": "configuration",
         "scope": "service",
         "value": "valueXXXX"
},
 "zzzz": {
        "type": "string",
         "description": "header zzzz for the service",
         "class": "header",
         "scope": "service",
         "value": "valueZZZZ"
},
"aaaa": {
        "type": "string",
         "description": "configuration aaaa for the service endpoint",
[...]

PR #211, opened 2025-03-26
 #185 Add the ability to override global Service configurat…

luigidemasi:
…ocation using Tool definition

### DIFF
--- a/core/core-exchange/src/main/java/ai/wanaku/core/exchange/ParsedToolInvokeRequest.java
+++ b/core/core-exchange/src/main/java/ai/wanaku/core/exchange/ParsedToolInvokeRequest.java
@@ -2,6 +2,7 @@ package ai.wanaku.core.exchange;
 
 
 import ai.wanaku.core.uri.Parameter;
+import ai.wanaku.core.uri.URIHelper;
 import ai.wanaku.core.uri.URIParser;
 import java.util.Collections;
 import java.util.HashMap;
@@ -10,7 +11,7 @@ import java.util.Map;
 /**
  * Represents a parsed tool invocation request containing the URI and its body.
  */
-public record ParsedToolInvokeRequest(String uri, String body) {
+public record ParsedToolInvokeRequest(String uri, String body, Map<String, String> headers) {
 
     /**
      * Parses the URI provided by the router
@@ -41,8 +42,13 @@ public record ParsedToolInvokeRequest(String uri, String body) {
 
         String parsedUri = URIParser.parse(uri, map);
 
+        //Add additional configuration
+        parsedUri = URIHelper.addQueryParameters(parsedUri, toolInvokeRequest.getServiceConfigurationsMap());
+
         String body = toolInvokeRequest.getBody();
 
-        return new ParsedToolInvokeRequest(parsedUri, body);
+        Map<String, String> headers = toolInvokeRequest.getHeadersMap();
+
+        return new ParsedToolInvokeRequest(parsedUri, body, headers);
     }
 }
--- a/core/core-util/src/main/java/ai/wanaku/core/uri/URIHelper.java
+++ b/core/core-util/src/main/java/ai/wanaku/core/uri/URIHelper.java
@@ -1,5 +1,6 @@
 package ai.wanaku.core.uri;
 
+
 import java.util.Map;
 
 /**
@@ -16,7 +17,10 @@ public class URIHelper {
     }
 
     private static void buildQuery(StringBuilder uri, Map<String, ?> parameters) {
-        boolean first = true;
+        buildQuery(uri, parameters, true);
+    }
+
+    private static void buildQuery(StringBuilder uri, Map<String, ?> parameters, boolean first) {
         for (Map.Entry<String, ?> entry : parameters.entrySet()) {
             if (first) {
                 uri.append('?');
@@ -65,5 +69,15 @@ public class URIHelper {
 
         uri.append(path);
         return buildFromBaseUri(uri, parameters);
+    }
+
+    public static String addQueryParameters(String uri, Map<String, ?> parameters) {
+        boolean first = true;
+        if (uri.contains("?")) {
+            first = false;
+        }
+        StringBuilder uriBuilder = new StringBuilder(uri);
+        buildQuery(uriBuilder, parameters, first);
+        return uriBuilder.toString();
     }
 }
--- a/routers/wanaku-router/src/main/java/ai/wanaku/routers/proxies/tools/InvokerProxy.java
+++ b/routers/wanaku-router/src/main/java/ai/wanaku/routers/proxies/tools/InvokerProxy.java
@@ -12,7 +12,6 @@ import ai.wanaku.core.exchange.ToolInvokeRequest;
 import ai.wanaku.core.exchange.ToolInvokerGrpc;
 import ai.wanaku.core.mcp.providers.ServiceRegistry;
 import ai.wanaku.core.util.CollectionsHelper;
-import ai.wanaku.core.util.ReservedArgumentNames;
 import ai.wanaku.routers.proxies.ToolsProxy;
 import com.google.protobuf.ProtocolStringList;
 import io.grpc.ManagedChannel;
@@ -23,7 +22,15 @@ import io.quarkiverse.mcp.server.ToolResponse;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
+
 import org.jboss.logging.Logger;
+
+import static ai.wanaku.core.util.ReservedArgumentNames.BODY;
+import static ai.wanaku.core.util.ReservedPropertyNames.SCOPE_SERVICE;
+import static ai.wanaku.core.util.ReservedPropertyNames.SCOPE_SERVICE_ENDPOINT;
+import static ai.wanaku.core.util.ReservedPropertyNames.TARGET_CONFIGURATION;
+import static ai.wanaku.core.util.ReservedPropertyNames.TARGET_HEADER;
 
 /**
  * A proxy class for invoking tools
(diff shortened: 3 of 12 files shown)


core/core-exchange/src/main/java/ai/wanaku/core/exchange/ParsedToolInvokeRequest.java
```diff
@@ -41,8 +42,13 @@ public static ParsedToolInvokeRequest parseRequest(String uri, ToolInvokeRequest
 
         String parsedUri = URIParser.parse(uri, map);
 
+        //Add additional configuration
+        parsedUri = URIHelper.addQueryParameters(parsedUri, toolInvokeRequest.getServiceConfigurationsMap());
+
```
orpiske, line 47: So, I'm thinking we should have this as an optional post processing step. I believe that not all services will want to have to append service configurations to the URI (specially, those that are not Camel-based). So, I think we should have a `Function<String, String> postProcessor` as an argument, that can receives the URI, does something like `parsedUri = postProcessor.accept(parsedUri)`. This would give more flexibility for those services ... 

To be clear: I don't think you need to change this now. I'm just dumping a thought that came to me while reviewing this ...

core/core-util/src/main/java/ai/wanaku/core/uri/URIHelper.java
```diff
@@ -66,4 +70,12 @@ public static String buildUri(String scheme, String path, Map<String, ?> paramet
         uri.append(path);
         return buildFromBaseUri(uri, parameters);
     }
+
+    public static String addQueryParameters(String uri, Map<String, ?> parameters) {
+        boolean first = true;
+        if (uri.contains("?")) {first = false;}
```
orpiske: Small nitpick: make sure to use multilines for the if-else here. It makes it easier to read the code.

routers/wanaku-router/src/main/java/ai/wanaku/routers/proxies/tools/InvokerProxy.java
```diff
@@ -72,12 +74,43 @@ private static ToolInvokeReply invokeRemotely(
...
+                .getInputSchema()
+                .getProperties();
+
+        Map<String,String> serviceEndpointConfiguration = inputSchema.entrySet()
+                .stream()
+                .filter(entry -> {
+                    ToolReference.Property property = entry.getValue();
+                    return property !=null &&
+                    property.getTarget().equals("configuration") &&
+                    property.getScope().equals("service-endpoint");
```
orpiske: Let's use a constant for these and for the `header` and the `service` strings below. It makes it easier for us to document this.

routers/wanaku-router/src/main/resources/application.properties
```diff
@@ -7,6 +7,7 @@ quarkus.log.category."ai.wanaku".level=DEBUG
 quarkus.log.category."io.quarkiverse.mcp".level=INFO
 quarkus.log.category."io.quarkiverse.quinoa".level=WARN
 
+quarkus.http.host=0.0.0.0
```
orpiske: Out of scope, no?

orpiske (changes requested):
This is much better than what we have today. For the implementation I only have a few very small nitpick items. 

When I run the tests, I did notice that the code is failing:

```
(.venv) [~/code/python/test-agent]$ python3 basic-test.py
Using prompt: Please give me 3 dog facts
---------- user ----------
'Please give me 3 dog facts'
/Users/opiske/code/python/test-agent/.venv/lib/python3.13/site-packages/autogen_agentchat/agents/_assistant_agent.py:870: UserWarning: Finish reason mismatch: stop != tool_calls when tool_calls are present. Finish reason may not be accurate. This may be due to the API used that is not returning the correct finish reason.
  model_result = await model_client.create(
---------- tester_assistant ----------
[FunctionCall(id='0', arguments='{"count": "3"}', name='dog-facts')]
---------- tester_assistant ----------
[FunctionExecutionResult(content='Error: unhandled errors in a TaskGroup (1 sub-exception)', name='dog-facts', call_id='0', is_error=True)]
---------- tester_assistant ----------
There is no tool available for this request.
Using prompt: Please give me 2 cat facts
---------- user ----------
'Please give me 2 cat facts'
---------- tester_assistant ----------
[FunctionCall(id='1', arguments='{"count": "2"}', name='meow-facts')]
---------- tester_assistant ----------
[FunctionExecutionResult(content='Error: unhandled errors in a TaskGroup (1 sub-exception)', name='meow-facts', call_id='1', is_error=True)]
---------- tester_assistant ----------
There is no tool available for this request.
```

Then, in the router logs I have this:

```
2025-03-27 09:42:31,766 ERROR [ai.wan.rou.pro.too.InvokerProxy] (vert.x-worker-thread-1) Unable to call endpoint: Cannot invoke "String.equals(Object)" because the return value of "ai.wanaku.api.types.ToolReference$Property.getTarget()" is null

Exception in InvokerProxy.java:86
          84                      ToolReference.Property property = entry.getValue();
          85                      return property !=null &&
        → 86                      property.getTarget().equals("configuration") &&
          87                      property.getScope().equals("service-endpoint");
          88                  })

Exception in InvokerProxy.java:89
          87                      property.getScope().equals("service-endpoint");
          88                  })
        → 89                  .collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().getValue()));
          90  
          91          //override the global configuration with the invocation specific one.

Exception in InvokerProxy.java:52
          50          LOG.infof("Invoking %s on %s", toolReference.getType(), service);
          51          try {
        → 52              final ToolInvokeReply invokeReply = invokeRemotely(toolReference, toolArguments, service);
          53  
          54              if (invokeReply.getIsError()) {: java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because the return value of "ai.wanaku.api.types.ToolReference$Property.getTarget()" is null
        at ai.wanaku.routers.proxies.tools.InvokerProxy.lambda$invokeRemotely$0(InvokerProxy.java:86)
        at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:178)
        at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
        at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1939)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
        at ai.wanaku.routers.proxies.tools.InvokerProxy.invokeRemotely(InvokerProxy.java:89)
        at ai.wanaku.routers.proxies.tools.InvokerProxy.call(InvokerProxy.java:52)
        at ai.wanaku.server.quarkus.api.v1.tools.ToolsBean.lambda$registerTool$1(ToolsBean.java:60)
        at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$FeatureDefinitionInfoBase.call(FeatureManagerBase.java:425)
        at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$1.call(FeatureManagerBase.java:70)
        at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$1.call(FeatureManagerBase.java:67)
        at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$ActivationSupport.call(FeatureManagerBase.java:269)
        at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$3.call(FeatureManagerBase.java:205)
        at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$3.call(FeatureManagerBase.java:201)
        at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$4(ContextImpl.java:192)
        at io.vertx.core.impl.ContextInternal.dispatch(ContextInternal.java:270)
        at io.vertx.core.impl.ContextImpl$1.execute(ContextImpl.java:221)
        at io.vertx.core.impl.WorkerTask.run(WorkerTask.java:56)
        at org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
        at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2675)
        at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2654)
        at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1627)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1594)
        at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:11)
        at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:11)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:1583)

2025-03-27 09:42:31,767 INFO  [io.qua.mcp.ser.traffic] (vert.x-worker-thread-1) MCP message sent [ZTYzOTliMTItZDk2ZS00Mjc0LWIxNDAtMDhmMjVmN2MxNDRl]:

{
  "jsonrpc" : "2.0",
  "id" : 1,
  "result" : {
    "isError" : true,
    "content" : [ {
      "text" : "Cannot invoke \"String.equals(Object)\" because the return value of \"ai.wanaku.api.types.ToolReference$Property.getTarget()\" is null",
      "type" : "text"
    } ]
  }
}
```

Maybe I need to recreate the storage files?

In any case, this is looking good ... looking forward to having this one in the code.

core/core-exchange/src/main/java/ai/wanaku/core/exchange/ParsedToolInvokeRequest.java
luigidemasi: @orpiske I like the idea of post processor functions, but we cannot pass it in the configuration, it should be provided by the service. There should be a sort of trigger that fires the post processor execution functions, in the case above, something like  predicate: 

`if target == 'configuration' and scope == 'service-endpoint'`


WDYT?

orpiske: Yeah, sounds good to me. 

Obs.: we can leave this for later ... I'm mostly thinking about a future improvement (and also the fact that we need to consider non-Camel workloads).

orpiske (approved):
Looking good, thanks @luigidemasi!

PR #213, opened 2025-03-28
Make naming more consistent

orpiske:


orpiske:
I'm going to merge this soon, but it's a bit disruptive as I am renaming several services, packages and classes so that we have consistent naming @Croway @luigidemasi @lordrip @ibek. Once merged, make sure to remove the files in `$HOME/.m2/repository/ai/wanaku` and then rebuild the code.

PR #219, opened 2025-03-31
Documentation updates

orpiske:


docs/usage.md
```diff
@@ -269,8 +269,22 @@ open http://localhost:8080
...
+Wanaku works with [HyperChat](https://github.com/BigSweetPotatoStudio/HyperChat). To do so,
+you can configure Wanaku as an MCP server using the MCP configuration as shown below:
+
+![HyperChat Configuration](imgs/hyperchat-configuration.png)
+
+> [!IMPORTANT]
+> Make sure to have Wanaku up and running before configuring HyperChat. You may also need to 
+> close and reopen HyperChat.
+
+After configuring HyperChat, you need to add or edit the agent and mark Wanaku as an allowed MCP server.
```
llowinge: @orpiske `edit the agent and mark Wanaku as an allowed MCP server` where ?

orpiske: I adjusted a bit, wdyt?

llowinge: It is now more clear. Thank you, last question - do i need the agent ? It looks i can chat even without creating the agent.

orpiske: So, I've just tried without an agent and ... it seems it works as well. It's kinda confusing, TBH. Let me rephrase it to sound a bit more optional

llowinge: Yeah, it is cause by my lack of knowledge. I'm not sure what the agent should suppose to do in this context.

PR #225, opened 2025-04-01
Cleanup old persistence API usages

orpiske:
- Remove unused interfaces related to the old persistence API
- Do not query for files directly in the router

### DIFF
--- a/core/core-mcp/src/main/java/ai/wanaku/core/mcp/common/resolvers/Resolver.java
+++ b/core/core-mcp/src/main/java/ai/wanaku/core/mcp/common/resolvers/Resolver.java
@@ -1,28 +1,16 @@
 package ai.wanaku.core.mcp.common.resolvers;
 
-import java.io.File;
-
 /**
  * A resolver that consumes MCP requests and resolves what type of tool or resource acquirer
  * should handle it
  */
 public interface Resolver {
+
+    // Whatever is using this, needs to get it from the persistence API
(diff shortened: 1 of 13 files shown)


orpiske:
c/c: @Croway

orpiske:
This removes some code still relying on the old persistence and makes sure that all persistence checks/initialization etc should happen via the new persistence API.

core/core-mcp/src/main/java/ai/wanaku/core/mcp/common/resolvers/Resolver.java
```diff
@@ -1,28 +1,16 @@
 package ai.wanaku.core.mcp.common.resolvers;
 
-import java.io.File;
-
 /**
  * A resolver that consumes MCP requests and resolves what type of tool or resource acquirer
  * should handle it
  */
 public interface Resolver {
-    String DEFAULT_RESOURCES_INDEX_FILE_NAME = "resources.json";
```
Croway, line 10: omg, good catch, I had this stuff in another branch and totally forgot about it!

Thanks, I hope it didn't require a lot of refactor

orpiske: No, it was a pretty quick refactor ... I noticed while working on the MCP forward and fixed it rather quickly.

PR #226, opened 2025-04-01
[closes #215] Add JBang support

### LINKED ISSUE #215: Install wanaku CLI with JBang
Camel has `jbang app install camel@apache/camel`, we co do the similar.

llowinge:
Ensure you have local 0.0.4-SNAPSHOT build, then you can try to run it with 

```
jbang run -Dwanaku.version=0.0.4-SNAPSHOT wanaku@llowinge/wanaku/jbang-support start local
```
If you will be OK with this approach, i'll also add some basic readme how to run it.

### DIFF
--- a/jbang-catalog.json
+++ b/jbang-catalog.json
@@ -1,0 +1,12 @@
+{
+  "catalogs": {},
+  "aliases": {
+    "wanaku": {
+      "script-ref": "jbang/src/main/java/ai/wanaku/jbang/WanakuJBang.java",
+      "dependencies": [
+        "ai.wanaku:cli:${wanaku.version:0.0.3}"
+      ],
+      "description": "Run Wanaku easily"
+    }
+  }
+}
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -25,12 +25,30 @@ The MCP endpoint exposed by Wanaku can be accessed on the path `/mcp/sse` of the
 locally, that would mean `http://localhost:8080/mcp/sse`)
 
 ## Getting the CLI
+### Getting the CLI by downloading binary
 
 The best way to install the CLI is by downloading the latest `cli` from the [latest release](https://github.com/wanaku-ai/wanaku/releases).
 
 > [!NOTE]
 > You may also find a container for the CLI on our [Quay.io organization](https://quay.io/repository/wanaku/cli), 
 > although it **is not entirely tested** at the moment.
+
+### Getting the CLI via JBang
+First, you must install [JBang](https://www.jbang.dev/). See instructions on [JBang](https://www.jbang.dev/download/) how to download and install.
+
+After JBang is installed, you can verify JBang is working by executing the following command from a command shell:
+```shell
+jbang version
+```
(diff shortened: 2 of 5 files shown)


llowinge:
@orpiske @Croway Please review, thank you.

orpiske (approved):
LGTM! Thanks @llowinge

orpiske:
I've just tried with ` jbang app install -Dwanaku.version=0.0.4-SNAPSHOT wanaku@llowinge/wanaku/jbang-support` 

And it works nicely well!!!

jbang-catalog.json
```diff
@@ -0,0 +1,12 @@
+{
+  "catalogs": {},
+  "aliases": {
+    "wanaku": {
+      "script-ref": "jbang/src/main/java/ai/wanaku/jbang/WanakuJBang.java",
+      "dependencies": [
+        "ai.wanaku:cli:${wanaku.version:0.0.3}"
```
llowinge, line 7: @orpiske This would be good to be updated with each release, right ? For that, i need to understand more, how is the release done now. But i guess we can live with manual bumping for some early time.

orpiske: We can use a resource filter like this:

1. Adjust the pom: https://github.com/wanaku-ai/wanaku/blob/main/core/core-util/pom.xml#L50-L58
2. Replace the fixed version with `${project.version}`, like here: https://github.com/wanaku-ai/wanaku/blob/main/core/core-util/src/main/resources/version.txt

orpiske: Or just leave it empty, like we do it in Camel. I think that way it picks the latest published version. I can publish the jars for 0.0.4-SNAPSHOT as soon as I have some time, so it would work until we have a release.

llowinge: @orpiske I guess we want to declare default value to only final versions (not snapshots), but right now, `project.version` is snapshot. I thought we would like to update it only to new final versions, together with release procedure.

orpiske: Yeah, makes sense. It would be good to find a way to do that automatically, as the release process itself is already a bit complicated. I think it's OK for now, but would be good to find a way to automate this.

Croway, line 7: why is it needed? can't you add the dependency here https://github.com/wanaku-ai/wanaku/pull/226/files#diff-74d36cc231abb4699b2017fbb0a2c0b452925d4ddcbb65c9d1b63cd92da876f7R7 ? as in Camel https://github.com/apache/camel/blob/main/dsl/camel-jbang/camel-jbang-main/dist/CamelJBang.java#L23

llowinge: I somehow liked it more this way, so we don't track dependency versions in Java file (i don't like mixing concerns too much). That is the only reason, i could move it to the Java file if you prefer.

Croway: okok! sounds good, no problem

Croway (approved):

llowinge:
Docs should be read @orpiske, i hope it is OK.

Issue #62, opened 2025-02-26
MCP as a tool or resource provider


orpiske:
We need to be able to use other downstream MCP resources and tools as tools/resources themselves. We should probably add a MCP consumer on Camel using [LangChain4j MCP client](https://docs.langchain4j.dev/tutorials/mcp/) and then use that to enable MCP downstreams.

orpiske:
Created [CAMEL-21825](https://issues.apache.org/jira/browse/CAMEL-21825) on Camel's Jira for this.

orpiske:
Decided to go via a different route as the MCP client cannot be added to Camel and we need greater integration within wanaku for this.

orpiske:
🎉 This issue has been resolved in `v0.0.5` ([Release Notes](https://github.com/wanaku-ai/wanaku/releases/tag/v0.0.5))

PR #233, opened 2025-04-03
Add a MCP-to-MCP bridge

orpiske:
With this, we integrate directly with MCP without a more complex approach relying on #62.


orpiske:
There's still quite a few things to do, but sharing it early in case anyone wants to start playing with it.

PR #236, opened 2025-04-04
Added a new API for updating resources, tools and forwards

### LINKED ISSUE #162: Update and override existing tool
I need an API/UI option to update and override existing tools. Currently, reimporting a toolset fails if tools already exist, requiring manual removal first. I want an option to override existing tools with the new specifications during import

orpiske:
This closes #162


orpiske:
@Croway please, would you mind a second look on the persistent part?

Croway (approved):

Croway:
> @Croway please, would you mind a second look on the persistent part?

I think that for now, the update is fine like this, once wanaku will be more stable, we can implement more performant updates (and a sql persistence implementation)

Issue #143, opened 2025-03-10
Toolset generator


orpiske:
Some systems may have very large sets of APIs. For those, generating tools and resources manually is simply not practical. We should investigate whether we can generate the toolsets automatically from their OpenAPI contract (idea originally shared by @ibek)

ibek:
Here's my experimentation with a customGPT https://chatgpt.com/g/g-67d0adfde19481919c88ee4c12660417-wanaku-toolset  ... however those APIs which I used are not fully working due to #155

ibek:
Here are the instructions: You are a developer specializing in transforming OpenAPI contracts and other API implementations into a toolset compatible with the Wanaku MCP (Model Context Protocol) Router. Your primary function is to analyze API specifications and convert them into JSON arrays of objects that conform to the Wanaku MCP Router’s format.

You must always adhere to the following structure format:
[{
  "name" : "example-name",
  "description" : "Description of the API",
  "uri" : "https://example.com/api",
  "type" : "http",
  "inputSchema" : {
    "type" : "object",
    "properties" : { },
    "required" : null
  }
}]

All properties defined in the "inputSchema" must be incorporated into the "uri" field, either as part of the URL path (e.g., "https://example.com/api/{param}") or as query parameters (e.g., "https://example.com/api?param={param}"). The placement should match the API’s expected format.

Your responses must be concise, structured, and formatted correctly. If additional details are required to complete a toolset entry, you ask specific clarifying questions before generating the output. You prioritize correctness, consistency, and maintainability in every response.

orpiske:
This is blocked by #155 (done) and #185 (in progress).

orpiske:
🎉 This issue has been resolved in `v0.0.6` ([Release Notes](https://github.com/wanaku-ai/wanaku/releases/tag/v0.0.6))

PR #241, opened 2025-04-07
issue #143 : Toolset generator

luigidemasi:


cli/src/main/java/ai/wanaku/cli/main/commands/tools/ToolsGenerate.java
```diff
@@ -0,0 +1,71 @@
...
+public class ToolsGenerate implements Callable<Integer> {
+
+    private static final Logger LOG = Logger.getLogger(ToolsGenerate.class);
+
+
+
+    @CommandLine.Option(names = { "-h", "--help" }, usageHelp = true, description = "Display the help and sub-commands")
+    private boolean helpRequested = false;
+
+    @CommandLine.Option(names = {"--serverUrl", "-u"}, description = "Override the OpenAPI server base url", arity = "0..1")
```
orpiske: For consistency, let's name the options with dash to separate words. For instance `--server-url`.

cli/src/main/java/ai/wanaku/cli/main/commands/tools/ToolsGenerate.java
```diff
@@ -0,0 +1,71 @@
...
+    protected String serverUrl;
+
+    @CommandLine.Option(names = {"--outputFile", "-o"}, description = "path of the file where the toolset will be written. If not set, the toolset will be written to the STDOUT", arity = "0..1")
```
orpiske: Same note about parameter names here and on the other options below.

cli/src/main/java/ai/wanaku/cli/main/support/CollectionHelper.java
```diff
@@ -0,0 +1,48 @@
+package ai.wanaku.cli.main.support;
+
+import java.util.Collection;
+import java.util.Map;
+
+public class CollectionHelper {
```
orpiske: We could merge this one with `core/core-util/src/main/java/ai/wanaku/core/util/CollectionsHelper.java`.

cli/src/main/java/ai/wanaku/cli/main/support/FileHelper.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private FileHelper() {}
+
+    public static boolean canWriteToDirectory(Path dir) {
+        if (dir == null) {
+            return false;
+        }
+
+        try {
+            Path tempFile = Files.createTempFile(dir, "test", ".tmp");
```
orpiske: Maybe let's pick a more appropriate prefix here. Let's say ... `wanaku-tmp` instead of `test`.

cli/src/main/java/ai/wanaku/cli/main/support/FileHelper.java
```diff
@@ -0,0 +1,28 @@
...
+            Files.delete(tempFile);
+            return true;
+        } catch (IOException e) {
```
orpiske, line 19: Maybe log here.

cli/pom.xml
```diff
@@ -91,6 +111,35 @@
...
+      <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-mockito</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.5.3</version>
```
orpiske: Let's use a property here for consistency ... We can move this to the `parent/pom.xml` file.

orpiske (changes requested):
This is looking very good! There's only a few minor things to adjust and I think we can get it in 0.0.4.

orpiske:
Ah, one more thing: we need to include a brief documentation about this in the usage-guide.

cli/src/main/java/ai/wanaku/cli/main/support/FileHelper.java
luigidemasi: This is just an helper method to check if a directory is writable, does it make sense to log something?

orpiske: Ah, got it. No worries, then.

orpiske (approved):

orpiske:
Thanks @luigidemasi, this looks great!

PR #280, opened 2025-04-30
Fix legacy GRPC and system module warnings

zbendhiba:


orpiske:
Thanks! Last time I tried this, it did not work as no gRPC server was listening. I'll test again soon.

PR #330, opened 2025-06-06
Fix a bug that prevents running wanaku using java -jar

Croway:


orpiske:
LGTM, but ... I'm a bit lost. What was happening?

Issue #29, opened 2025-02-22
Increase test coverage


orpiske:
At large our test coverage is very poor: unit and integration tests should be added for pretty much all of the features.

PR #333, opened 2025-06-06
Add MCP Inspector Junit Extension

Croway:


test-base/mcp-inspector/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+        <dependencies>
+            <dependency>
+                <groupId>${quarkus.platform.group-id}</groupId>
+                <artifactId>${quarkus.platform.artifact-id}</artifactId>
+                <version>${quarkus.platform.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
orpiske: This one should come from the `parent/pom.xml` file: https://github.com/wanaku-ai/wanaku/blob/main/parent/pom.xml

pom.xml
```diff
@@ -105,6 +105,7 @@
         <module>services</module>
         <module>archetypes</module>
         <module>jbang</module>
+        <module>test-base</module>
```
orpiske: I'm thinking that maybe this one should be build first, so we can use this test module on the code.

orpiske (changes requested):
This is looking really nice @Croway!

I made a few comments, but mostly cosmetic/small stuff. 

You can also annotate the commit with a reference to #29 as this is somewhat related to that one. 

Thanks!

PR #339, opened 2025-06-10
Notifications targets endpoint and UI handling

Croway:
@lordrip @ibek @orpiske 

This is a first attempt, it works, but I am not an SSE expert (especially on the UI side), any comments is welcome.

There are some TODOs on the code @orpiske maybe you can tell if those makes sense or not, moreover I was not able to generate an Update event, therefore.. we'll test it in production :)


orpiske (approved):
This looks pretty nice! Thank you @Croway !

PR #346, opened 2025-06-12
Add Edit Tool command

luigidemasi:


core/core-persistence/core-persistence-infinispan/src/main/java/ai/wanaku/core/persistence/infinispan/InfinispanToolReferenceRepository.java
```diff
@@ -26,4 +30,13 @@ protected Class<ToolReference> entityType() {
...
         return UUID.randomUUID().toString();
-    }}
+    }
+
+    @Override
+    public List<ToolReference> findByName(String name) {
+      Query<ToolReference> query = cacheManager.getCache(entityName()).query("from ai.wanaku.api.types.ToolReference t where t.name = :name");
+      query.setParameter("name", name);
+      return query.execute().list();
+    }
```
orpiske, line 40: I like this. I wasn't aware we could do it. There's a few places where we might benefit from this approach.

luigidemasi: Yeah, that's why I like Infinispan, you get a cache with a SQL-like query capabilities, two birds with one stone.

lordrip:
> It's failing because of UI build issues. Please @Croway and/or @lordrip, would you guys have a sec to let us know what could be the problem here?

Here you have: https://github.com/luigidemasi/wanaku/pull/1

cc @luigidemasi

Issue #347, opened 2025-06-13
Add the possibility to list the service capabilities  from CLI


luigidemasi:
`wanaku services list`

luigidemasi:
yes, exactly!

orpiske:
> yes, exactly!

Nice, +1 from me!

orpiske:
🎉 This issue has been resolved in `v0.0.6` ([Release Notes](https://github.com/wanaku-ai/wanaku/releases/tag/v0.0.6))

PR #369, opened 2025-06-24
Add the possibility to list the service capabilities from CLI #347

luigidemasi:


orpiske:
Man, this looks nice! Pretty cool stuff ... But I think it may be impacted by #363? As in, we need to adjust this, so that it shows the configurations what is displayed comes from the config URIs (i.e.: `file://my-sqs.properties`)

orpiske:
> @orpiske sure, I can adjust it once it will be merged #363. Meanwhile can we deprecate targets?

+1 for deprecating targets! The capabilities feature you provided here is much cleaner.

cli/pom.xml
```diff
@@ -82,6 +82,11 @@
                 <version>${juniversalchardet-version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>io.quarkiverse.langchain4j</groupId>
+                <artifactId>quarkus-langchain4j-mcp</artifactId>
+            </dependency>
+
```
orpiske: This one

PR #389, opened 2025-06-30
Create an MCP servers module

### LINKED ISSUE #377: Create an MCP servers module
For adding MPC servers to use as standalone MCP or integrated in Wanaku, like for example

https://github.com/oscerd/aws-security-tool

oscerd:
Fixes #377


mcp-servers/aws-security-tool/aws-security-quarkus.log
orpiske, line 1: I guess we don't need this one, right?

mcp-servers/aws-security-tool/mvnw
orpiske, line 1: I think we also don't need the wrapper

mcp-servers/aws-security-tool/src/main/java/com/github/oscerd/S3Tool.java
```diff
@@ -0,0 +1,192 @@
+package com.github.oscerd;
```
orpiske: I think we can change this to `ai.wanaku.mcp.servers.aws.security`

orpiske (commented):
It looks good to me. I think you just need to remove the wrapper, the log and adjust the package. Later we can create an archetype to create these more easily and also leverage the #379 (but I'll test that one after we merge this ... it's the perfect one for me to try it)

oscerd:
I think it's ok now.

PR #398, opened 2025-07-02
Namespaces UI

Croway:
This PR add the following:

- new Page Namespaces, that shows the available namespace on wanaku
image
- add namespace select in Tools/Resources modal and handle the namespace in the submit
- show the namespace in the Tools/Resources table
image

Improvements (@lordrip I need your input here) the namespaces can be fetched and cached at the beginning, the namespaces do not change often. do you have any pointer what's the best practice to achieve it?

Future improvements, instead of showing the namespace in the Tools/Resources table, create a global select that filter the namespaces (like openshift).


ui/src/hooks/api/use-namespaces.ts
```diff
@@ -0,0 +1,18 @@
...
+        ): Promise<getApiV1NamespacesListResponse> => {
+          return getApiV1NamespacesList(options);
+        },
+        []
+    ); 
+
+    return {
+        listNamespaces
+    };
+}
```
lordrip, line 45: [friendly-comment] This could be a bit misleading because there's no cache here, and depending on how we use it, it will end up calling the namespaces endpoint on every render.

```tsx
const Component = () => {
    const { listNamespaces } = useNamespaces();
    namespaces = listNamespaces();

    // Use namespaces for something

    return <span>Hola</span>
}
```

An alternative might be using a UI store and fetching the namespaces at the beginning of the application and from that point, use it in the other components. Alternatively, if we want to continue to be simple, having a parent component, fetching the namespaces and passing them down to the component that needs to rerender might be another option.


```tsx
const NamespacePage = () => {
    const { listNamespaces } = useNamespaces();
    const namespaces = useMemo(() => listNamespaces(), [listNamespaces]);

    return <NamespaceComponent namespaces={namespaces} />
}

const NamespaceComponent = ({ namespaces }) => {
    // use namespaces for something here

    return <span>Hola</span>
}
```

ui/src/Pages/Tools/ToolsPage.tsx
```diff
@@ -147,6 +148,19 @@ const AddToolModal: React.FC<AddToolModalProps> = ({
...
   const [inputSchema, setInputSchema] = useState("");
+  const { listNamespaces } = useNamespaces();
+  const [fetchedNamespaceData, setFetchedNamespaceData] = useState<Namespace[]>([]);
+  const [selectedNamespace, setSelectedNamespace] = useState('');
+    
+    useEffect(() => {
+      listNamespaces().then((result) => {
+      setFetchedNamespaceData(result.data.data as Namespace[]);
+      });
+    }, [listNamespaces]);
```
lordrip, line 161: I think this brings a bit of a flash in the UI, How about fetching the namespaces with a loading spinner and then applying it to the UI?

Croway: I would like to load the namespace on the first load and cache it.

Are there best practices for use cases like this?

orpiske:
> Hi @lordrip @orpiske @ibek I just pushed a 2nd attempt, I've been reading about caches, but I could not find any best practice about this topic, there are sooo many libraries. And well, when in doubt, I just used a singleton that works as expected. any feedback is welcome. but I achieved what I wanted, the namespaces are loaded only the first time, and then cached

I think not having a cache is fine, for now. Caching is hard ... Also, I'd probably prefer to have it cached on the backend, so it could also serve the CLI in a consistent manner. In any case, I think it's fine to leave it for later (unless there is a significant impact on the UI)

orpiske:
I guess this one is ready for merging, right?

Croway:
@orpiske @ibek the Types select are now loaded dynamically, based on what downstream service is currently running

orpiske:
> @orpiske @ibek the Types select are now loaded dynamically, based on what downstream service is currently running

Thank you!!!

PR #434, opened 2025-07-16
Added support for using the integration tests locally

orpiske:
This makes it possible to run manual tests with the builtin infrastructure

## Summary by Sourcery

Enable local execution of integration tests by extracting CLI invocation logic into a helper class, updating existing tests to use it, and adding manual-only integration tests with exclusion from default test runs.

New Features:
- Introduce CLIHelper to execute Wanaku CLI commands with a configurable host parameter
- Add WanakuHttpToolManualIT for manual local integration testing

Enhancements:
- Refactor executeWanakuCliCommand out of the base test class and update integration tests to use the new helper

Build:
- Exclude **ManualIT** tests from the default Maven test execution in the integration pom

Tests:
- Update WanakuFileResourceIT and WanakuHttpToolIT to call CLIHelper.executeWanakuCliCommand with the host argument

### DIFF
--- a/tests/integration/src/test/java/ai/wanaku/mcp/CLIHelper.java
+++ b/tests/integration/src/test/java/ai/wanaku/mcp/CLIHelper.java
@@ -1,0 +1,56 @@
+package ai.wanaku.mcp;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.List;
+import org.assertj.core.api.Assertions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import picocli.CommandLine;
+
+public class CLIHelper {
+    private static final Logger LOG = LoggerFactory.getLogger(CLIHelper.class);
+
+    private CLIHelper() {}
+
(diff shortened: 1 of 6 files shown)


orpiske:
@Croway with this, we can test without having to push containers. This makes debugging easier ...

tests/integration/src/test/java/ai/wanaku/mcp/CLIHelper.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param host The host running the Wanaku MCP Router
+     * @return The exit code of the command.
+     */
+    public static int executeWanakuCliCommand(List<String> command, String host) {
+        List<String> executableCommand = new ArrayList<>(command);
+        if ("wanaku".equals(executableCommand.get(0))) {
+            executableCommand.remove(0);
+        }
+
+        executableCommand.add(String.format("--host=%s", host));
```
Croway: host may be null here, it should be set only if host != null

orpiske: Aha, got it.

PR #476, opened 2025-08-21
Add a CLI option for creating a new mcp services.

VaishnaviR18:


cli/src/main/java/ai/wanaku/cli/main/support/WanakuCliConfig.java
```diff
@@ -17,10 +17,16 @@ interface Resource {
         String createCmd();
     }
 
+    interface Mcp {
+        String createCmd();
+    }
```
orpiske, line 22: I think it is missing the configuration on the `application.properties` file.

orpiske (changes requested):
LGTM, but I think it is missing the configuration on the `application.properties'.

cli/src/main/java/ai/wanaku/cli/main/support/WanakuCliConfig.java
VaishnaviR18: @orpiske Sorry, I missed it.
I have added it, can you please review? Thanks!

PR #487, opened 2025-09-17
Upgrade Quinoa to the latest version (2.6.2)

orpiske:
## Summary by Sourcery

Upgrade Quinoa and related tooling, refresh generated API clients, update package manager configurations, and clean up API specification settings.

Enhancements:
- Upgrade Quarkus Quinoa integration to v2.6.2 with updated Yarn, Node and added pnpm versions
- Update UI package manager to yarn@1.22.22 and refresh lockfile
- Regenerate API client models with orval v7.11.2 and switch response types from void to null
- Remove unused OpenID Connect security scheme from the OpenAPI spec
- Enable and expand CORS configuration in the backend to include localhost:8080


orpiske:
Screenshot 2025-09-17 at 17 11 41

This is failing due to CORS (c/c: @Croway )

orpiske:
Ok. It seems to have worked.

wanaku-router-backend/src/main/resources/application.properties
```diff
@@ -21,8 +21,8 @@ quarkus.log.level=INFO
 quarkus.log.category."ai.wanaku".level=DEBUG
 quarkus.log.category."io.quarkiverse.mcp".level=DEBUG
 
-%dev.quarkus.http.cors.enabled=true
-%dev.quarkus.http.cors.origins=http://localhost:8081,http://127.0.0.1:8081,http://localhost:5173
+quarkus.http.cors.enabled=true
+quarkus.http.cors.origins=http://localhost:8081,http://127.0.0.1:8081,http://localhost:5173,http://localhost:8080
```
Croway, line 25: @orpiske this is why I didn't notice this issue. When starting the web module with `quarkus:dev`, I was using `http://localhost:5173` therefore, no CORS error, I wonder why `http://localhost:8080` is needed :/

orpiske: Yeah, and that looks wrong, tbh ... At least this gives me some time to research the problem and understand why it needs it

PR #489, opened 2025-09-18
Enable authentication by default

orpiske:
## Summary by Sourcery

Enable authentication by default and restructure the router into a single backend module with integrated SPA build, update deployment configurations and documentation to reflect the new setup

New Features:
- Enable OpenID Connect authentication by default for router and core capability services
- Integrate Quarkus Quinoa into the router backend to build and serve the frontend SPA

Enhancements:
- Unify router components under a new wanaku-router aggregator and retire the standalone router-web module
- Update Docker Compose and OpenShift manifests to point services at wanaku-router-backend and configure default auth flags
- Simplify the CLI local runner to only wait for the router backend startup and remove web UI startup logic
- Consolidate POM module structure, JReleaser distributions, and project references to reflect the new router aggregator

Build:
- Add quarkus-quinoa dependency and configuration in the router backend POM for SPA packaging

Documentation:
- Update usage documentation to reflect the new UI port (8080), note auth defaults, and remove outdated router-web instructions


ui/src/custom-fetch.ts
```diff
@@ -8,9 +8,38 @@ const getBody = <T>(c: Response | Request): Promise<T> => {
...
+      console.log('Error fetching backend URL, using origin:', cachedBackendUrl, error);
+    }
+
+    return cachedBackendUrl!;
+  };
+
   // NOTE: Update just base url
   export const getUrl = (contextUrl: string): string => {
-    const baseUrl = VITE_API_URL || window.location.origin;
+    const baseUrl = getBackendUrl();
```
orpiske: Please, I need your thoughts on this change here @lordrip @Croway. 

Long story short: the router was split into 2 parts for the authentication support. One is the web and another is the router backend. 

The challenge is that the the web needs a way to know how/where to find the backend, so it can issue the requests to that address. This can be dynamic (i.e.: for instance, when running on k8s).

To solve the problem, I created an endpoint that returns the configured address for the server (i.e.: provided by the BackendResource. Then, I modified this code to query that endpoint once to get the router address, cache the value and use it for subsequent calls.

Very likely not my brightest technical work, so I'd like to hear your thoughts about this ...

Croway: mhhhh this sounds a bit hacky to me. 

I'd use `.env.*` standards. We can have two (or more) .env.local or .env.development files with the environemnt variable `VITE_API_URL=http://localhost:8080` if the backend is running somewhere else, the user will have to change it. But I think this is pretty standard in front end development. In order to do so, the vite config has to be updated and the property `quarkus.quinoa.package-manager-command.dev-env.VITE_API_URL=\"http://localhost:8080\"` has to be removed.

wdyt @lordrip ?

lordrip: I'm not sure I understand the problem yet. From where is the UI being served?

Croway: a new java module `wanaku-router-web` serves the UI (via `mvn quarkus:dev` for example). And this is the resource added by Otavio https://github.com/wanaku-ai/wanaku/pull/489/files#diff-30613ebd61eb2e2ba3f7dbe8740438b62a68d0ca585f65ac2ad98864660d08d3R22 that returns the actual backend address.. To me it sounds something like a BFF (back end for frontend), but I do not think we need that. That's why I was suggesting to follow frontend best practices with `.env` files.. but again, I am not an expert in this field

orpiske: So, I had a quick chat with @lordrip about this approach and as you both pointed out, this is not the best approach. Let me get back to the drawing board and rethink this approach.

PR #513, opened 2025-10-06
Fixed integration tests to work with authentication

orpiske:


orpiske:
Just doing the basics here, as most of these will be affected by a major refactoring of the capabilities soon.

PR #632, opened 2025-11-03
Add label support with expression filtering and batch operations

### LINKED ISSUE #628: Add label support with expression filtering and batch operations
Adds comprehensive label support to the Wanaku CLI, enabling users to organize and manage tools, resources, and forwards using key-value labels with powerful filtering and batch operations.

luigidemasi:
resolve #628


cli/src/main/java/ai/wanaku/cli/main/support/WanakuPrinter.java
```diff
@@ -311,6 +317,103 @@ public void printSuccessMessage(String message) {
         }
     }
 
+    /**
+     * Prints Markdown text with terminal styling and formatting.
+     *
+     * <p>Parses and renders Markdown text with ANSI color codes and text formatting
```
orpiske, line 352: Just a heads up that these `<p>` and `</p>` tend to cause the Javadoc generation to fail. I'd recommend doing a dist build to make sure none of these cause problems.

core/core-mcp/src/main/java/ai/wanaku/core/mcp/util/LabelExpressionParser.java
```diff
@@ -0,0 +1,308 @@
...
+ * <pre>
+ * expression  ::= orExpr
+ * orExpr      ::= andExpr ('|' andExpr)*
+ * andExpr     ::= notExpr ('&' notExpr)*
+ * notExpr     ::= '!' notExpr | primary
+ * primary     ::= '(' expression ')' | comparison
+ * comparison  ::= IDENTIFIER '=' STRING | IDENTIFIER '!=' STRING
+ * </pre>
+ */
+public class LabelExpressionParser {
```
orpiske, line 30: It would be good to have extra testing coverage on this one.

cli/src/main/resources/docs/LABEL_EXPRESSIONS.md
```diff
@@ -0,0 +1,463 @@
+# Label Expressions Guide
```
orpiske, line 1: I'm wondering if there's a way we could have the content of this file both as part of the `man` command, but also online and indexable by search engines.

orpiske (commented):
Nice!

core/core-mcp/src/main/java/ai/wanaku/core/mcp/util/LabelExpressionParser.java
luigidemasi: Do you think https://github.com/wanaku-ai/wanaku/pull/632/files#diff-8fdf81d83d084e57c90f539a72e60c9782c5a55964badf8b7f4795012e12626c is not enough?

orpiske: I missed that one. TBH, it's a bit confusing that we have the `LabelExpressionParser` in one module and the test in a different one. I would be better if both could be on the same one.

core/core-persistence/core-persistence-api/src/main/java/ai/wanaku/core/persistence/api/WanakuRepository.java
```diff
@@ -88,4 +88,6 @@ public interface WanakuRepository<A extends WanakuEntity, C> {
      * @throws IllegalArgumentException if fields map is null or empty
      */
     int removeByFields(Map<String, Object> fields);
+
+    int removeAll();
```
orpiske, line 98: It would be good to document this one.

core/core-services-api/src/main/java/ai/wanaku/core/services/api/CapabilitiesService.java
```diff
@@ -71,12 +81,22 @@ Response toolsConfigure(
...
      * Lists all available resource capability providers.
      *
+     * @param labelFilter optional label expression to filter resources by labels
      * @return a {@link WanakuResponse} containing a list of resource service targets
      */
     @Path("/resources/list")
     @GET
     @Consumes(MediaType.TEXT_PLAIN)
-    WanakuResponse<List<ServiceTarget>> resourcesList();
+    WanakuResponse<List<ServiceTarget>> resourcesList(@QueryParam("labelFilter") String labelFilter);
```
orpiske, line 90: I see option to list filtering capabilities, but I don't see where labels are associated with capabilities.

orpiske (changes requested):
Overall it LGTM, however I noticed that it does not seem to provide a way to associate labels to capabilities (that is OK - I am not sure if we need to do that). However, it offers a way to list based on labels.

Therefore, I think we either need to allow capabilities to associate labels to themselves (which may be complicated) or it needs to not allow filtering capabilities based on labels.

cli/src/main/java/ai/wanaku/cli/main/commands/capabilities/CapabilitiesList.java
```diff
@@ -71,6 +71,18 @@ public class CapabilitiesList extends BaseCommand {
...
+            names = {"-e", "--label-expression"},
+            description = {
+                """
+Filter capabilities by label expression. Supports logical operators for complex queries.
+For detailed information see the label expression manual page:
+`wanaku man label-expression`
+Note: If omitted, all capabilities are listed. Label matching is case-sensitive.
+"""
+            })
+    private String labelExpression;
```
orpiske, line 84: We can probably leave this aside for now, until we decide whether we want to assign labels to capabilities

orpiske (changes requested):
I think the only thing that needs adjustments is having the labels on the client-side features, but not on the service-side (i.e.: forwards, capabilities, etc) as we need to investigate how that would affect how they are matched on each case.  

Overall, it LGTM.

PR #634, opened 2025-11-05
Add support for prompts.

VaishnaviR18:
## Summary by Sourcery

Implement end-to-end support for reusable MCP prompts including schema definitions, persistence, backend service and REST endpoints, CLI commands, and UI pages with documentation and tests.

New Features:
- Add full CRUD support for MCP prompts via API, CLI, and UI.

Enhancements:
- Extend OpenAPI schema and generated client/models with prompt schemas and endpoints
- Integrate prompt management into backend with PromptManager registration, template expansion, and persistence via Infinispan
- Implement Protostream marshallers for prompt-related types and register serialization schemas

Build:
- Add jackson-annotations dependency to API module

Documentation:
- Add detailed documentation for managing prompts in CLI and UI in docs/usage.md and update contributing guide

Tests:
- Add integration tests for prompts REST resource and persistence tests for PromptReference repository


core/core-services-api/pom.xml
```diff
@@ -16,6 +16,11 @@
             <artifactId>jakarta.ws.rs-api</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.eclipse.microprofile.rest.client</groupId>
+            <artifactId>microprofile-rest-client-api</artifactId>
+        </dependency>
```
orpiske: I think this one should not be a part of this implementation. The API should not depend on anything else other than jakarta.

VaishnaviR18: Ohhh okay!
I will remove this and will rebase and adjust the code.
Thank you!

core/core-persistence/core-persistence-infinispan/src/main/resources/proto/types.proto
luigidemasi, line 1: the proto file needs to be splitted, one file per message. take a look at the proto files in the main branch.

core/core-services-api/src/main/java/ai/wanaku/core/services/api/PromptsService.java
```diff
@@ -0,0 +1,44 @@
...
+import jakarta.ws.rs.QueryParam;
+import jakarta.ws.rs.core.MediaType;
+import java.util.List;
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+
+@RegisterRestClient
+@Path("/api/v1/prompts")
+public interface PromptsService {
+
+    @Path("/add")
```
luigidemasi: I'm not a huge fan of putting the action name in the path. Let's use the HTTP verb instead

core/core-services-api/src/main/java/ai/wanaku/core/services/api/PromptsService.java
```diff
@@ -0,0 +1,44 @@
...
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    WanakuResponse<PromptReference> add(PromptReference prompt);
+
+    @Path("/list")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    WanakuResponse<List<PromptReference>> list();
+
+    @Path("/remove")
```
luigidemasi: use DELETE Http verb and remove the action from the path

core/core-services-api/src/main/java/ai/wanaku/core/services/api/PromptsService.java
```diff
@@ -0,0 +1,44 @@
...
+    @PUT
+    void remove(@QueryParam("prompt") String promptName);
+
+    @Path("/update")
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    void update(PromptReference prompt);
+
+    @Path("/")
```
luigidemasi:
```java
    @Path("/")
    @GET
    @Produce(MediaType.APPLICATION_JSON)
    WanakuResponse<PromptReference> getByName(@QueryParam("name") String name);
```

core/core-persistence/core-persistence-infinispan/src/main/resources/proto/types.proto
VaishnaviR18: Thank you, i will update it.

core/core-services-api/src/main/java/ai/wanaku/core/services/api/PromptsService.java
VaishnaviR18: Okay! Thank you!

core/core-services-api/src/main/java/ai/wanaku/core/services/api/PromptsService.java
VaishnaviR18: Okay, thanks!

core/core-services-api/src/main/java/ai/wanaku/core/services/api/PromptsService.java
orpiske: That's OK to me, but it's important to be consistent. So, we can use the verbs, but we also need to ensure that the API is consistent across the back-end. I created https://github.com/wanaku-ai/wanaku/issues/636 for this.

cli/src/main/java/ai/wanaku/cli/main/commands/prompts/PromptsAdd.java
```diff
@@ -0,0 +1,238 @@
...
+            names = {"--configuration-from-file"},
+            description = "Configure the prompt using the given file",
+            arity = "0..1")
+    private String configurationFromFile;
+
+    @CommandLine.Option(
+            names = {"--secrets-from-file"},
+            description = "Add the given secrets to the prompt using the given file",
+            arity = "0..1")
+    private String secretsFromFile;
```
orpiske: I don't think we need these. Both are used by the capabilities, on the gRPC bridge to provision configurations and secrets when adding tools.

wanaku-router/wanaku-router-backend/src/main/java/ai/wanaku/backend/resolvers/WanakuPromptsResolver.java
```diff
@@ -0,0 +1,34 @@
...
+public class WanakuPromptsResolver implements PromptsResolver {
+
+    @Inject
+    PromptReferenceRepository promptReferenceRepository;
+
+    @Override
+    public void provision(PromptPayload promptPayload) throws ServiceNotFoundException {
+        // For now, prompts don't require provisioning from external services
+        // This could be extended in the future if prompts need to fetch data from services
+    }
```
orpiske: I think we won't need this. Or, at least, not in the same way we have provisioning for add tools/resources. The reason being that, if capabilities were to provide prompts, then I believe they would provide it on connection. We can think a bit more about this when we start working on sampling and elicitation.

wanaku-router/wanaku-router-backend/src/main/java/ai/wanaku/backend/common/PromptHelper.java
```diff
@@ -0,0 +1,164 @@
...
+
+        io.quarkiverse.mcp.server.Content mcpContent;
+
+        if (content instanceof ai.wanaku.api.types.TextContent) {
+            // Text content
+            ai.wanaku.api.types.TextContent textContent = (ai.wanaku.api.types.TextContent) content;
+            mcpContent = new TextContent(textContent.getText());
+        } else if (content instanceof ImageContent) {
+            // Image content (base64 encoded data)
+            ImageContent imageContent = (ImageContent) content;
```
orpiske: Just a minor nitpick here, please use the new (i.e.: Java 17+) pattern for doing this: ```content instanceof ImageContent imageContent```. This is cleaner to read.

api/pom.xml
```diff
@@ -12,4 +12,11 @@
     <artifactId>api</artifactId>
     <name>Wanaku :: API</name>
 
+    <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+        </dependency>
+    </dependencies>
```
orpiske, line 20: This dependency is unused. It should not be part of the API module.

orpiske (changes requested):
Thanks, it's almost there ... I think the unused dependency on the `api` needs to be removed. Then, it should be ready to merge.

orpiske (approved):
Vaishnavi explained about the reason for the dependency. We should be good to go. Thanks!

Issue #928, opened 2026-03-30
Add CLI commands to configure MCP clients (Claude Desktop, Cursor, etc.)


orpiske:
## Description

Add `wanaku configure` CLI commands that help users quickly set up popular MCP clients to connect to their Wanaku server.

Example usage:
```
wanaku configure claude          # Configure Claude Desktop to connect to wanaku server
wanaku configure cursor --transport http --port 8080  # Configure Cursor to connect to local wanaku server
```

This would streamline the onboarding experience by automating MCP client configuration, reducing manual setup steps for users.

## Expected Behavior

- `wanaku configure <client>` generates/updates the appropriate configuration file for the specified MCP client
- Supports common options like `--transport` and `--port`
- Initially targets Claude Desktop and Cursor, with an extensible design for adding more clients

## Additional Context

Inspired by similar functionality in other MCP tooling. The goal is to make it as easy as possible for users to connect their preferred MCP clients to a Wanaku server.

PR #937, opened 2026-03-31
feat: CLI commands to configure MCP clients Claude Desktop, Cursor

matheusandre1:
Closes: #928 

In this case, ConfigureMcpClientCommand should be an abstract, right? Adopting a strategy pattern would be a bit complex for now, I believe.

## Summary by Sourcery

Add a new CLI `configure` command with subcommands to set up MCP clients for Wanaku, including Claude Desktop and Cursor, and document their usage.

New Features:
- Introduce a `configure` CLI command group with subcommands for configuring Claude Desktop and Cursor MCP clients to connect to Wanaku.
- Support configuration options for MCP transport, host, port, and server name when generating client configuration files.

Enhancements:
- Factor out shared MCP client configuration logic into an abstract base command to reuse validation and config file handling across clients.

Documentation:
- Update usage documentation to describe the new `wanaku configure claude` and `wanaku configure cursor` commands and how they modify MCP client configuration files.

Tests:
- Add tests covering configuration generation for Claude Desktop and Cursor, including merging existing configs and handling unsupported transports.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/configure/ConfigureClaude.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/configure/ConfigureClaude.java
@@ -47,11 +47,13 @@ public class ConfigureClaude extends ConfigureMcpClientCommand {
     protected ObjectNode createServerEntry(URI endpoint) {
         ObjectNode entry = newObjectNode();
         entry.put("command", "uvx");
-        if ("http".equalsIgnoreCase(transport == null ? null : transport.trim())) {
         if ("http".equalsIgnoreCase(transport.trim())) {
-                    .add("mcp-proxy")
-                    .add("--transport=streamablehttp")
-                    .add(endpoint.toString()));
+            entry.set(
+                    "args",
+                    newArrayNode()
+                            .add("mcp-proxy")
+                            .add("--transport=streamablehttp")
+                            .add(endpoint.toString()));
         } else {
             entry.set("args", newArrayNode().add("mcp-proxy").add(endpoint.toString()));
         }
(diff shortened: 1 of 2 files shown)


apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/configure/ConfigureClaude.java
```diff
@@ -0,0 +1,61 @@
...
+
+        return resolveOverrideOrDefault(home.resolve(".config").resolve("Claude").resolve("claude_desktop_config.json"));
+    }
+
+    @Override
+    protected ObjectNode createServerEntry(URI endpoint) {
+        ObjectNode entry = newObjectNode();
+        entry.put("command", "uvx");
+        if ("http".equalsIgnoreCase(transport == null ? null : transport.trim())) {
+            entry.set("args", newArrayNode()
```
orpiske: Nit: the null-safe check `transport == null ? null : transport.trim()` is unnecessary here — `transport` always has a default value of `"sse"` via picocli, and `resolveWanakuEndpoint()` in the base class validates it before `createServerEntry` is ever called. The guard adds confusion without value.

```suggestion
        if ("http".equalsIgnoreCase(transport.trim())) {
```

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/configure/ConfigureMcpClientCommand.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    protected URI resolveWanakuEndpoint() {
+        String normalizedTransport = transport == null ? "" : transport.trim().toLowerCase();
+        String path;
+        switch (normalizedTransport) {
+            case "sse":
+                path = "/mcp/sse/";
+                break;
```
orpiske, line 85: Consider marking this method `final` to match the other helper methods (`readOrCreateConfig`, `ensureObjectNode`, etc.) and signal that subclasses should not override the endpoint resolution logic.

```suggestion
    protected final URI resolveWanakuEndpoint() {
```

PR #961, opened 2026-04-06
fix: Zip Slip path traversal vulnerability in ZipHelper

matheusandre1:
Fixes: #954

## Summary by Sourcery

Address a Zip Slip path traversal vulnerability in the CLI ZipHelper and add regression tests to ensure extracted files remain within the intended component directory.

Bug Fixes:
- Prevent ZipHelper from extracting ZIP entries outside the target component directory by validating normalized extraction paths against the component base path.

Tests:
- Add unit tests covering normal extraction, flat and nested paths, and multiple path traversal attempts using both backslash and forward slash separators.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/ZipHelper.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/ZipHelper.java
@@ -6,6 +6,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Enumeration;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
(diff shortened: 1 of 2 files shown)


orpiske (changes requested):
On the right track, but would be good to take a look at those minor notes.

orpiske (approved):
## Review Summary

Both previous review notes have been addressed:

1. **Flat file edge case** — Handled with the `idx >= 0 && idx + 1 < entryName.length()` guard, and covered by the new `unzipKeepsFlatFilesInsideTheComponentDirectory` test.
2. **Forward-slash traversal test** — Added as `unzipRejectsForwardSlashPathTraversalEntries`.

The Zip Slip fix follows OWASP/Snyk best practices (`Path.normalize()` + `startsWith(componentPath)`). Test coverage is thorough with 6 test cases. The pre-existing `return` vs `continue` bug in the directory branch is also fixed. No blocking issues.

PR #964, opened 2026-04-08
refactor: GCache gRPC channels instead of recreating them on every request and add unit tests

### LINKED ISSUE #849: Cache gRPC channels instead of recreating them on every request
## Description

The gRPC channel is currently recreated on every request. This adds unnecessary overhead and can impact performance under load. Caching the channel and reusing it for subsequent requests to the same target would reduce connection setup costs and improve throughput.

## Expected Behavior

gRPC channels should be cached and reused for requests targeting the same service endpoint, rather than being created and torn down on every request.

matheusandre1:
Closes: #849

## Summary by Sourcery

Refactor gRPC channel management to cache and reuse channels per endpoint and add tests to verify the new lifecycle and reuse behavior.

Enhancements:
- Cache gRPC channels by endpoint string instead of ServiceTarget instances and reuse them across requests.
- Make request-level closeChannel a no-op while providing a shutdown method to close and clear all cached channels.

Tests:
- Add unit tests covering channel reuse across equivalent targets, creation for different endpoints, no-op close behavior, and shutdown semantics including idempotency and empty cache handling.

### DIFF
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/transports/grpc/GrpcChannelManager.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/transports/grpc/GrpcChannelManager.java
@@ -16,41 +16,29 @@ import ai.wanaku.capabilities.sdk.api.types.providers.ServiceTarget;
  * features like connection pooling, interceptors, and custom configurations
  * in the future.
  * <p>
- * Current implementation creates channels with plaintext communication.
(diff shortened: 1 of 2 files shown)


orpiske (changes requested):
Thanks for trying this one @matheusandre1, but it is a bit more complex than that, so I can't accept this PR. 

This work is something I have on my pipeline for one of the upcoming versions (likely reusing the code from another project I am working on).

PR #983, opened 2026-04-16
Rethrowing exception as before in forwards registration

rnetuka:
Trying to fix exception handling from https://github.com/wanaku-ai/wanaku/pull/979

@orpiske What do you think?

## Summary by Sourcery

Adjust forward registration error handling to propagate underlying exceptions while mapping refresh failures to HTTP 500 responses.

Bug Fixes:
- Restore rethrowing of the original exception from forward registration instead of wrapping it in a WanakuException.
- Ensure the forwards refresh API responds with an HTTP 500 status when an unexpected exception occurs during refresh.

Enhancements:
- Declare checked exceptions on forward registration and refresh methods to reflect their error behavior.


orpiske:
Thanks @rnetuka ... I think this one here would do the trick: https://github.com/wanaku-ai/wanaku/pull/980

orpiske:
Thanks for this, but I ended up merging #980 as I also wanted to include a cleanup

PR #989, opened 2026-04-17
Bugfix: forwards look up namespace by id

rnetuka:
The namespace for forwards is passed by id from the UI (always from a select box). Namespace allocation is no longer needed when adding forwards (in fact, it was causing a bug to me to save the forward with _different_ namespace than expected).

Defaults to **default** namespace for extra safety

## Summary by Sourcery

Adjust forward namespace resolution to use namespace IDs and fall back to the default namespace when none is provided.

Bug Fixes:
- Resolve forwards being saved under an unexpected namespace by using the namespace ID supplied by the UI instead of allocating by name.
- Prevent null or empty forward namespaces from causing inconsistent behavior by defaulting to the 'default' namespace.

Enhancements:
- Add a helper in NamespacesBean to retrieve the configured default namespace, failing fast if none exists.


orpiske (approved):
LGTM, thanks!

PR #991, opened 2026-04-17
Bugfix: storing addresses instead of whole ForwardClients in the registry

rnetuka:
There was a resource leak in MCPClients / ForwardClients. The ForwardClients were stored in a registry with possibly opened connections. The latest AutoCloseable interface fix solved this, but the issue manifested itself in not listing any remote resources (the connection was already closed by that time).

Storing addresses instead of the whole clients in the ForwardRegistry. Create new clients on demand and close them once done working.

## Summary by Sourcery

Store only forward client addresses in the registry and create short-lived ForwardClient instances on demand when accessing or cleaning up forwards.

Bug Fixes:
- Fix resource leak and stale closed connections caused by storing long-lived ForwardClient instances in the registry.

Enhancements:
- Update forward registry and API bean to work with client addresses, creating and closing ForwardClient instances per operation instead of sharing them globally.


orpiske (approved):
LGTM, thanks

PR #1021, opened 2026-04-29
[UI] Enhancement: Import toolset modal overhaul

rnetuka:
For better user experience, I've enhanced the modal dialog for importing toolsets. 

The modal dialog is not two-step with the first step for providing remote URL for fetching a toolset. In second step, the fetched tools are listed and the user have an option to exclude them from the set before saving.

There is a possibility to skip the first step and copy-paste raw Json data. Also, if the tools were fetched from the remote server, it is still posible to edit them before saving if there is ever a need for it.

On top of that, I've added some (although limited) input validation both for the URL and the Json data.

import-toolset-1

import-toolset-2

import-toolset-3

## Summary by Sourcery

Overhaul the import toolset modal into a multi-step, validated workflow that supports fetching, reviewing, editing, and selectively importing tools.

New Features:
- Introduce a two-step import workflow that optionally fetches a toolset from a remote URL before review and import.
- Add a table-based list view for fetched tools that allows selecting which tools to include in the imported toolset.
- Allow switching between a list view and raw JSON view when preparing the toolset for import.

Bug Fixes:
- Improve robustness of toolset parsing by validating tool names and supporting both single-tool and array JSON payloads.

Enhancements:
- Add client-side validation for toolset URLs and JSON input, with surfaced error messages in the modal.
- Disable the final import action until valid toolset JSON is available, reducing the chance of invalid submissions.

### DIFF
--- a/apps/ui/admin/src/Pages/Tools/ImportToolsetModal.tsx
+++ b/apps/ui/admin/src/Pages/Tools/ImportToolsetModal.tsx
@@ -1,11 +1,17 @@
 import {
+  Column,
+  ContentSwitcher,
+  Grid,
   Modal,
   Stack,
+  Switch,
   TextArea,
   TextInput
 } from "@carbon/react"
-import {useState} from "react"
+import React, {useRef, useState} from "react"
 import {ToolReference} from "../../models"
+import {ToolParserError, Tools} from "./tools"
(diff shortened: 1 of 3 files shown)


orpiske (approved):
Man, this looks a lot nicer ... thanks!

PR #1032, opened 2026-04-30
Add support for configurable Keycloak realm in authentication settings

### LINKED ISSUE #1028: Support configurable authentication realms
## Description

The Keycloak realm name is currently hardcoded to `wanaku` in several places across the codebase. For example, in the operator's `OperatorUtil.java`, the token endpoint is built as:

```java
authServer + "/realms/wanaku"
```

This prevents users from integrating Wanaku with existing Keycloak deployments that use different realm names.

## Expected Behavior

Users should be able to configure the authentication realm name, rather than being locked into the `wanaku` realm. This should be supported across the operator CRD, CLI, and capabilities SDK.

## Additional Context

- There is an existing TODO in `OperatorUtil.java` (line 404): "this one should be made more flexible, so it can accept different realms"
- The `wanaku admin realm create` CLI command also assumes the `wanaku` realm name
- The CIC also hardcodes `/realms/wanaku` in its token endpoint construction

matheusandre1:
Closes: #1028

## Summary by Sourcery

Make the Keycloak realm used for authentication configurable across the operator, router backend, SDK, deployment scripts, and tests while keeping `wanaku` as the default.

New Features:
- Allow Wanaku capabilities and router resources to specify a Keycloak realm via an `authRealm` field in their auth configuration.
- Introduce configurable `auth.realm`/`AUTH_REALM` properties for Quarkus OIDC client and router backend to control the realm used in auth server URLs.
- Enable perf/load test scripts and auth configuration scripts to work with a configurable Keycloak realm instead of a hard-coded one.

Enhancements:
- Derive the operator token endpoint realm from capability configuration via a new helper, falling back to the default realm when unset or blank.

Deployment:
- Update Keycloak configuration script to support a configurable realm when regenerating client secrets.

Documentation:
- Document the configurable Keycloak realm, its default value, and how to set it via properties or environment variables in usage and configuration guides.
- Update sample YAML manifests to demonstrate optional `authRealm` usage in capability and router specs.

Tests:
- Add unit tests for realm resolution logic in the operator utilities and for the new `authRealm` field behavior in `WanakuTypes`.
- Adjust performance test scripts to parameterize the Keycloak realm used in OIDC client configuration.

### DIFF
--- a/apps/wanaku-operator/src/main/java/ai/wanaku/operator/util/OperatorUtil.java
+++ b/apps/wanaku-operator/src/main/java/ai/wanaku/operator/util/OperatorUtil.java
@@ -124,6 +124,8 @@ public final class OperatorUtil {
             }
         }
 
+        String realm = resolveAuthRealm(resource);
+
         EnvVar authServerEnv = new EnvVarBuilder()
                 .withName(EnvironmentVariables.AUTH_SERVER)
                 .withValue(authServer)
@@ -132,10 +134,15 @@ public final class OperatorUtil {
                 .withName(EnvironmentVariables.AUTH_PROXY)
                 .withValue(authProxy)
                 .build();
+        EnvVar authRealmEnv = new EnvVarBuilder()
+                .withName(EnvironmentVariables.AUTH_REALM)
+                .withValue(realm)
+                .build();
 
         List<EnvVar> envVars = new java.util.ArrayList<>();
         envVars.add(authServerEnv);
         envVars.add(authProxyEnv);
+        envVars.add(authRealmEnv);
 
         final WanakuRouterSpec.RouterSpec routerSpec = resource.getSpec().getRouter();
 
@@ -401,10 +408,10 @@ public final class OperatorUtil {
         // Build the registration URI - use the router ref for service discovery
         String registrationUri = getInternalRegistrationUri(resource);
 
-        // TODO: this one should be made more flexible, so it can accept different realms
+        String realm = resolveAuthRealm(resource);
         EnvVar authServerEnv = new EnvVarBuilder()
                 .withName(EnvironmentVariables.CAMEL_INTEGRATION_CAPABILITY_TOKEN_ENDPOINT)
-                .withValue(authServer + "/realms/wanaku")
+                .withValue(authServer + "/realms/" + realm)
                 .build();
         EnvVar registrationUriEnv = new EnvVarBuilder()
                 .withName(EnvironmentVariables.CAMEL_INTEGRATION_CAPABILITY_REGISTRATION_URL)
@@ -559,5 +566,21 @@ public final class OperatorUtil {
         service.addOwnerReference(resource);
 
         return service;
+    }
+
+    static String resolveAuthRealm(WanakuCapability resource) {
+        if (resource == null || resource.getSpec() == null || resource.getSpec().getAuth() == null) {
+            return EnvironmentVariables.DEFAULT_AUTH_REALM;
+        }
+        String realm = resource.getSpec().getAuth().getAuthRealm();
+        return (realm == null || realm.isBlank()) ? EnvironmentVariables.DEFAULT_AUTH_REALM : realm;
+    }
+
+    static String resolveAuthRealm(WanakuRouter resource) {
+        if (resource == null || resource.getSpec() == null || resource.getSpec().getAuth() == null) {
+            return EnvironmentVariables.DEFAULT_AUTH_REALM;
+        }
+        String realm = resource.getSpec().getAuth().getAuthRealm();
+        return (realm == null || realm.isBlank()) ? EnvironmentVariables.DEFAULT_AUTH_REALM : realm;
     }
 }
--- a/apps/wanaku-router-backend/src/main/webui/openapi.json
+++ b/apps/wanaku-router-backend/src/main/webui/openapi.json
@@ -942,7 +942,7 @@
     "securitySchemes" : {
       "SecurityScheme" : {
         "type" : "openIdConnect",
-        "openIdConnectUrl" : "http://localhost:8543/realms/wanaku/.well-known/openid-configuration",
+        "openIdConnectUrl" : "http://localhost:8543/realms/${auth.realm}/.well-known/openid-configuration",
         "description" : "Authentication"
       }
     }
--- a/apps/wanaku-router-backend/src/main/webui/openapi.yaml
+++ b/apps/wanaku-router-backend/src/main/webui/openapi.yaml
@@ -633,7 +633,7 @@ components:
   securitySchemes:
     SecurityScheme:
       type: openIdConnect
-      openIdConnectUrl: http://localhost:8543/realms/wanaku/.well-known/openid-configuration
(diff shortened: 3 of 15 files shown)


apps/wanaku-operator/src/main/java/ai/wanaku/operator/util/OperatorUtil.java
```diff
@@ -560,4 +560,12 @@ public static Service makeCapabilityInternalService(
 
         return service;
     }
+
+    static String resolveAuthRealm(WanakuCapability resource) {
```
orpiske, line 571: **Router container is missing equivalent realm handling.**

This `resolveAuthRealm()` method only accepts `WanakuCapability`, which correctly handles the capability deployment path. However, the router deployment path (`setupBackendContainer()` around line 106) sets `AUTH_SERVER` and `AUTH_PROXY` on the router container but never sets `AUTH_REALM`.

Since the router backend's `application.properties` now resolves the realm via `${AUTH_REALM:wanaku}`, the operator also needs to pass this env var from the `WanakuRouter` CR's `auth.authRealm` field. Without this, operator-deployed routers will silently ignore any `authRealm` configured in the `WanakuRouter` CR.

A similar resolve-and-inject approach is needed in `setupBackendContainer()`, either by generalizing this method to accept both CRD types (since both use `WanakuTypes.AuthSpec`) or by adding a parallel resolver for `WanakuRouter`.

docs/configurations.md
```diff
@@ -344,11 +344,13 @@ wanaku.service.registration.uri=http://wanaku-router:8080
 wanaku.service.registration.interval=15s
 wanaku.service.registration.announce-address=my-custom-tool.example.com:9010
 
-quarkus.oidc-client.auth-server-url=https://keycloak.example.com/realms/wanaku
+quarkus.oidc-client.auth-server-url=https://keycloak.example.com/realms/${auth.realm}
```
orpiske: This example block now shows a Quarkus property placeholder (`${auth.realm}`), which may be confusing in user-facing documentation. Consider showing the resolved default value instead:

```suggestion
quarkus.oidc-client.auth-server-url=https://keycloak.example.com/realms/wanaku
```

The configurable realm note you added below (line 352) already explains how to customize it, so the example itself can keep the concrete default.

docs/configurations.md
```diff
@@ -367,7 +369,7 @@ wanaku.service.registration.uri=http://wanaku-router:8080
...
```
orpiske: Same as above — consider using the resolved default value here for clarity:

```suggestion
quarkus.oidc-client.auth-server-url=https://keycloak.example.com/realms/wanaku
```

orpiske (changes requested):
Thanks for working on this, @matheusandre1 — this is a good first step toward making the Keycloak realm configurable. The `resolveAuthRealm()` logic and the test coverage for the capability path are solid.

A few items need to be addressed before this can be merged:

### Findings

1. **Router container is missing `AUTH_REALM` env var (blocking)** — The router backend's `application.properties` now reads `${AUTH_REALM:wanaku}`, but `setupBackendContainer()` in `OperatorUtil.java` never sets this env var on the router container. This means operator-deployed routers will always fall back to the default `wanaku` realm, even if `authRealm` is set in the `WanakuRouter` CR spec. See inline comment for details.
2. **Documentation examples use raw property placeholders** — In `docs/configurations.md`, the example blocks now show `${auth.realm}` which is a Quarkus property placeholder. For user-facing documentation examples, it would be clearer to show the resolved default (`wanaku`) with a note explaining the `auth.realm` / `AUTH_REALM` configurability, similar to how it's already handled in `docs/usage.md`. See inline comments.
3. **Auto-generated OpenAPI files should be regenerated** — `openapi.json` and `openapi.yaml` still reference `realms/wanaku` in `openIdConnectUrl`. Please include a regeneration pass so they reflect the configurable realm. Only the realm reference change should be included — no other modifications.

Note: this review covers project conventions and code-level observations. It does not replace specialized review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

apps/wanaku-router-backend/src/main/webui/openapi.json
```diff
@@ -942,7 +942,7 @@
     "securitySchemes" : {
       "SecurityScheme" : {
         "type" : "openIdConnect",
-        "openIdConnectUrl" : "http://localhost:8543/realms/wanaku/.well-known/openid-configuration",
+        "openIdConnectUrl" : "http://localhost:8543/realms/${auth.realm}/.well-known/openid-configuration",
```
orpiske, line 945: This is a static JSON file served to the web UI — Quarkus property substitution (`${...}`) does not apply here. The literal string `${auth.realm}` will appear in the OpenAPI spec instead of the actual realm name.

Since this file represents the development-default OpenAPI spec, it should keep the hardcoded default:

```suggestion
        "openIdConnectUrl" : "http://localhost:8543/realms/wanaku/.well-known/openid-configuration",
```

apps/wanaku-router-backend/src/main/webui/openapi.yaml
```diff
@@ -633,7 +633,7 @@ components:
   securitySchemes:
     SecurityScheme:
       type: openIdConnect
-      openIdConnectUrl: http://localhost:8543/realms/wanaku/.well-known/openid-configuration
+      openIdConnectUrl: http://localhost:8543/realms/${auth.realm}/.well-known/openid-configuration
```
orpiske, line 636: Same issue as the JSON file — this is a static YAML file, so `${auth.realm}` won't be resolved.

```suggestion
      openIdConnectUrl: http://localhost:8543/realms/wanaku/.well-known/openid-configuration
```

orpiske (changes requested):
Thanks for tackling this, @matheusandre1 — nice work resolving the hardcoded realm across the operator, router backend, and capabilities SDK. The test coverage for `resolveAuthRealm` is thorough, and the `configure-auth.sh` validation is a good safety net.

A few items to address:

### Convention notes

- ~**Branch naming:** Project guidelines expect `ci-issue-1028` (see `project-guidelines.md`). Current branch is `feat-1028`.~
- **Commit message format:** Expected `Fix #1028: <brief description>` per project conventions.

### Scope notes

- The `wanaku admin realm create` CLI command still assumes the `wanaku` realm — this is fine for now and can be documented separately.
- The Camel Integration Capability (CIC) token endpoint also hardcodes `/realms/wanaku` — a separate issue has been created in that repository to track it.

orpiske:
I will also review this manually now.

orpiske:
For the Camel Integration Capability: https://github.com/wanaku-ai/camel-integration-capability/issues/86

orpiske (approved):
Thanks for another great PR @matheusandre1. This LGTM to me.

matheusandre1:
@orpiske , I was going to send the revision you just made.

orpiske:
> @orpiske , I was going to send the revision you just made.

Oh, my bad ... I have just tested it on my cluster and it was OK. Please, if you don't mind, feel free to send the remaining bits in a separate PR.

PR #1123, opened 2026-05-14
Fix #1122: open JAR filesystem for service template loading

### LINKED ISSUE #1122: ServiceTemplateInitializer fails with FileSystemNotFoundException when running from JAR
When the router is deployed in OpenShift (packaged as a JAR), the `ServiceTemplateInitializer` fails at startup with:

```
java.nio.file.FileSystemNotFoundException
    at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getFileSystem(ZipFileSystemProvider.java:156)
    at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getPath(ZipFileSystemProvider.java:142)
    at java.base/java.nio.file.Path.of(Path.java:209)
    at ai.wanaku.backend.api.v1.servicecatalog.ServiceTemplateInitializer.resolveTemplatesDir(ServiceTemplateInitializer.java:73)
```

`Path.of(resource.toURI())` does not work with `jar:` URIs because the ZIP filesystem for the JAR has not been opened. A `FileSystem` must be explicitly created via `FileSystems.newFileSystem()` before resolving paths inside the JAR.

orpiske:
## Summary
- Fix `FileSystemNotFoundException` when `ServiceTemplateInitializer` runs from a packaged JAR (e.g., OpenShift deployment)
- Detect `jar:` URI scheme and open a `FileSystem` via `FileSystems.newFileSystem()` before resolving template paths
- Extract deployment logic into `deployFromDir()` so the JAR filesystem can be properly scoped with try-with-resources
- Handle `FileSystemAlreadyExistsException` for dev-mode hot-reload scenarios by falling back to the existing filesystem

## Test plan
- [ ] Deploy router as a JAR in OpenShift and verify startup succeeds without `FileSystemNotFoundException`
- [ ] Verify built-in service templates are loaded correctly on startup
- [ ] Verify dev-mode (exploded classpath) still works as before
- [ ] Verify dev-mode live reload doesn't fail with `FileSystemAlreadyExistsException`

### DIFF
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/servicecatalog/ServiceTemplateInitializer.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/servicecatalog/ServiceTemplateInitializer.java
@@ -11,6 +11,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.DirectoryStream;
 import java.nio.file.FileSystem;
+import java.nio.file.FileSystemAlreadyExistsException;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


orpiske:
Thanks for the review feedback!

**1. Handle `FileSystemAlreadyExistsException`** — Applied in 59ff87cd0. The code now catches `FileSystemAlreadyExistsException` and falls back to `FileSystems.getFileSystem(uri)`, only closing the filesystem if we created it (to avoid closing one owned by another component during dev-mode reload).

**2. Derive internal path from the URI** — Declining this one. The resource is located via `getResource(TEMPLATES_RESOURCE)`, so the constant is guaranteed to match the path inside the JAR. Parsing the URI's scheme-specific part would introduce fragility (different URI formats across classloaders) without a practical benefit.

PR #1125, opened 2026-05-14
Fix #1005: make API key optional for local LLM providers

### LINKED ISSUE #1004: Make LLM chat base URL allowlist configurable via application properties
## Description

As part of PR #973 (`172543e536686287c49787f75126d0c40556363e`), the LLM chat backend was reorganized and a base URL allowlist was introduced. The allowlist is currently hardcoded in the backend implementation, which makes it difficult to manage across environments or extend without a code change.

In particular, `LlmChatResource` defines a static `List.of(...)` with the allowed chat endpoints and uses it both for `/api/v1/chat/allowlist` and to validate requests sent to `/api/v1/chat/completions`.

## Expected Behavior

The LLM chat allowlist should be configurable through application properties so operators can manage allowed chat providers per deployment.

A likely direction would be to externalize the values from `apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java` into a config property and keep the existing endpoint behavior unchanged.

## Actual Behavior

The allowlist is hardcoded in code:

- `apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java:34`
- Validation depends on the same static list at `apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java:58`

## Additional Context

Current hardcoded entries include:
- `http://localhost:11434`
- `https://api.openai.com`
- `https://api.mistral.ai`
- `https://generativelanguage.googleapis.com/v1beta/openai/`
- `https://api.anthropic.com`

This follow-up comes from the LLM chat overhaul in PR #973:
https://github.com/wanaku-ai/wanaku/pull/973

### LINKED ISSUE #1005: LLM chat fails when using local models without an API key
## Description

The LLM chat backend throws an exception when the selected model endpoint does not require an API key, such as a locally hosted Ollama instance.

`LlmChatResource` currently reads `apiKey` unconditionally from the request payload, which causes a failure when the field is absent or null.

## Steps to Reproduce

1. Open the LLM chat UI.
2. Select a local model endpoint such as `http://localhost:11434`.
3. Submit a chat request without providing an API key.
4. Observe the backend failure.

## Expected Behavior

Requests to local model providers that do not require authentication should succeed without requiring an API key.

## Actual Behavior

The backend throws an exception at:

- `apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java:55`

Observed exception:

```text
java.lang.NullPointerException: Cannot invoke "jakarta.json.JsonString.getString()" because the return value of "org.eclipse.parsson.JsonObjectBuilderImpl.getJsonString(String)" is null
```

## Additional Context

Relevant code path:

- `apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java:53`
- `apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java:55`
- `apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java:67`

The current implementation does:

- `String apiKey = json.getString("apiKey");`
- Always sets `Authorization: Bearer <apiKey>` on the outbound request

This follow-up likely affects locally hosted providers such as Ollama, where authentication is typically not required.

orpiske:
## Summary
- Make `apiKey` extraction null-safe in `LlmChatResource` using `json.containsKey()` before calling `getString()`
- Only add the `Authorization: Bearer` header when `apiKey` is non-null and non-empty
- Fixes `NullPointerException` when using local model endpoints (e.g., Ollama) that don't require authentication

## Test plan
- [x] Added `LlmChatResourceTest` with cases for missing apiKey, empty apiKey, and disallowed URL
- [x] Manual: verify LLM chat works with a local Ollama instance without providing an API key
- [x] Manual: verify LLM chat still works with OpenAI/Mistral when an API key is provided

## Summary by Sourcery

Allow chat completions to work without requiring an API key and add coverage for API key edge cases and URL allowlisting.

Bug Fixes:
- Avoid NullPointerException when apiKey is missing or null in LlmChatResource.
- Prevent sending an Authorization header when the apiKey is empty or absent.

Tests:
- Add LlmChatResourceTest covering missing and empty apiKey handling and rejection of disallowed base URLs.

### DIFF
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java
@@ -9,6 +9,8 @@ import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.WebApplicationException;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
 
 import java.io.IOException;
 import java.io.StringReader;
@@ -21,22 +23,22 @@ import java.net.http.HttpResponse;
 import java.net.http.HttpResponse.BodyHandlers;
 import java.util.List;
 
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
+
 import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
 import static jakarta.ws.rs.core.Response.Status.BAD_REQUEST;
 import static jakarta.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static jakarta.ws.rs.core.Response.Status.OK;
 import static jakarta.ws.rs.core.Response.Status.TOO_MANY_REQUESTS;
 
 @ApplicationScoped
 @Path("/api/v1/chat")
 public class LlmChatResource {
+    private static final Logger LOG = Logger.getLogger(LlmChatResource.class);
 
-    /* Allowlist of chat URLs */
-    private final List<String> allowlist = List.of(
-            "http://localhost:11434",
-            "https://api.openai.com",
-            "https://api.mistral.ai",
-            "https://generativelanguage.googleapis.com/v1beta/openai/",
-            "https://api.anthropic.com");
+    @ConfigProperty(name = "wanaku.chat.allowlist")
+    List<String> allowlist;
 
     @GET
     @Path("/allowlist")
@@ -52,7 +54,7 @@ public class LlmChatResource {
     public String codeCompletions(String data) {
         JsonObject json = Json.createReader(new StringReader(data)).readObject();
         String baseUrl = json.getString("baseUrl");
-        String apiKey = json.getString("apiKey");
+        String apiKey = json.containsKey("apiKey") ? json.getString("apiKey") : null;
         JsonObject llmParams = json.getJsonObject("chatParams");
 
         if (!allowlist.contains(baseUrl)) {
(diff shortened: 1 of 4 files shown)


apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java
claudio4j: Can you add a `LOG.errorf("HTTP Response Error: %s", response.body());` ?
So, whenever there is a 500 error, it's printed in the log.

claudio4j:
rebase from main to fix the `FirstAvailableTest.testResolveServiceNotFound`

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java
orpiske: Absolutely!

orpiske:
Added: moved the hardcoded chat LLM allowlist from `LlmChatResource` to the `wanaku.chat.allowlist` configuration property in `application.properties`. This makes it configurable via any Quarkus config source (env vars, config files, etc.).

orpiske:
> rebase from main to fix the `FirstAvailableTest.testResolveServiceNotFound`

Hm, I'm not seeing that one on `main`. Which one is it, please?

claudio4j:
> Hm, I'm not seeing that one on `main`. Which one is it, please?

Not failing anymore. Only NamespacesBeanTest is failing.

PR #1128, opened 2026-05-15
feat: URL allowlist for toolset repositories to mitigate SSRF risks

### LINKED ISSUE #1042: Add URL allowlist for toolset repositories to mitigate SSRF risks
## Description

The toolset repository feature (added in #1040) includes basic URL validation that blocks localhost and private IP ranges. However, a more robust approach would be to support a configurable URL allowlist so administrators can restrict which hosts/domains the backend is allowed to fetch from.

### Current state
- URLs are validated to require `http`/`https` scheme
- `localhost`, `127.0.0.1`, `::1`, `0.0.0.0`, and RFC 1918 private ranges are blocked

### Proposed enhancement
- Add a configurable allowlist (e.g., via `application.properties` or admin API)
- When an allowlist is configured, only URLs matching allowlisted host patterns should be accepted
- Ensure redirect handling does not bypass the allowlist
- Consider supporting glob patterns (e.g., `*.github.com`, `*.githubusercontent.com`)

### Context
Raised during code review of #1041 — see https://github.com/wanaku-ai/wanaku/pull/1041#discussion_r3180895493

matheusandre1:
Closes: #1042

## Summary by Sourcery

Introduce configurable URL allowlist and strengthened network address validation for toolset repository access to mitigate SSRF risks.

New Features:
- Add configurable glob-based URL allowlist for toolset repositories via the `wanaku.toolset-repos.url-allowlist` property.
- Expose the effective toolset repository URL allowlist through a new `/api/v1/toolset-repos/allowlist` REST endpoint.

Enhancements:
- Enhance URL validation for toolset repositories to cover loopback, private, link-local, carrier-grade NAT, and reserved IP ranges using DNS/InetAddress resolution and fallback patterns.
- Integrate URL validation into toolset repository browsing and fetching operations to enforce security checks before outbound requests.

Documentation:
- Document the toolset repository URL allowlist configuration, default SSRF protections, glob pattern semantics, and example configurations in the configuration guide.

Tests:
- Add unit tests for URL validation behavior across public, private, loopback, and special address ranges with and without an allowlist configured.
- Add unit tests for URL allowlist glob pattern translation and host matching, including configuration edge cases.


orpiske:
## 🔍 Issue Analysis: URL allowlist for toolset repositories (#1128)

### Summary

PR #1128 implements a configurable URL allowlist for toolset repository access, as requested in issue #1042. The feature adds glob-based host pattern matching, enhanced SSRF protections (loopback, private, link-local, CGNAT, reserved IPv6), a new REST endpoint for inspecting the allowlist, and comprehensive documentation and tests.

---

### Scope & Changes (671 additions, 9 deletions across 13 files)

**Core logic (3 files):**
- `ToolsetReposBean.java` — URL validation refactored from `static` to instance method; injected `UrlAllowlistConfig`; added `isLoopbackHost()`, `isPrivateOrReservedHost()`, `isCarrierGradeNat()`, `isReservedIpv6()`, `isPrivateIpv4Pattern()` helpers; validation now called in `browse()` and `fetchToolset()` code paths
- `UrlAllowlistConfig.java` — **New** `@ApplicationScoped` config bean; reads `wanaku.toolset-repos.url-allowlist` property; parses comma-separated glob patterns; lazy-compiled regex cache with `volatile` + double-checked locking
- `ToolsetReposResource.java` — New `GET /api/v1/toolset-repos/allowlist` endpoint

**Configuration & docs (3 files):**
- `application.properties` (main + test) — New `wanaku.toolset-repos.url-allowlist=` property
- `docs/configurations.md` — Documents allowlist behavior, glob semantics, and examples

**API spec (2 files):**
- `openapi.json` / `openapi.yaml` — New `WanakuResponseListString` schema and `/api/v1/toolset-repos/allowlist` endpoint

**UI (3 files):**
- `wanaku-router-api.ts`, `index.ts`, `wanakuResponseListString.ts` — Generated client types and fetch function for the new endpoint

**Tests (2 files):**
- `ToolsetReposBeanUrlValidationTest.java` — 197 lines; covers schemes, host presence, loopback, private, link-local, CGNAT, allowlist interaction
- `UrlAllowlistConfigTest.java` — 113 lines; covers glob-to-regex conversion, pattern matching, configuration edge cases

---

### Architecture Analysis

**Dual-mode validation design:**
1. **No allowlist configured (default):** Blocks loopback, private (RFC 1918), link-local (169.254.x), CGNAT (100.64–127.x), multicast, reserved IPv6 — all public URLs allowed
2. **Allowlist configured:** Only hosts matching glob patterns accepted; all others rejected regardless of public/private status

**Validation call sites:**
- `add()` — Already existed (validates on repo registration)
- `browse()` — **New** (validates before fetching `index.properties`)
- `fetchToolset()` — **New** (validates before fetching toolset JSON)
- `update()` — Already existed (validates URL on update if changed)

**Glob pattern semantics:**
| Pattern | Meaning |
|---|---|
| `*` | Matches a single subdomain label (no dots) |
| `**` | Matches zero or more subdomain levels |
| `?` | Matches a single character |

---

### Key Findings & Potential Issues

#### 1. 🔴 HTTP Redirect Bypass (SSRF Gap) — High Severity
The most significant gap identified: **URL validation only occurs before the initial HTTP request, not after redirects.** Both `ToolsetRepoIndex.fromUrl()` (using `URL.openStream()`) and `ToolsetIndexHelper.loadToolsIndex()` (using Jackson's `ObjectMapper.readValue(URL, ...)`) rely on Java's default `HttpURLConnection` which **follows HTTP 3xx redirects automatically**. This means:
- An attacker could register an allowlisted host (e.g., `api.github.com`) that returns a 302 redirect to `http://169.254.169.254/latest/meta-data/` (cloud metadata endpoint)
- The allowlist validation would pass for the initial URL, but the actual HTTP request would follow the redirect to the internal target

This was explicitly listed as a requirement in issue #1042 ("Ensure redirect handling does not bypass the allowlist") and is not addressed.

#### 2. 🟡 DNS Rebinding / TOCTOU — Medium Severity
`isLoopbackHost()` and `isPrivateOrReservedHost()` resolve the hostname via `InetAddress.getByName()` at validation time. However, `ToolsetRepoIndex.fromUrl()` opens a new connection later. Between validation and the actual connection, DNS records could change (DNS rebinding attack), pointing the same hostname to a different IP. This is a well-known TOCTOU (time-of-check-time-of-use) class of vulnerability.

#### 3. 🟡 Double-Checked Locking Pattern — Medium Concern
`UrlAllowlistConfig` uses `volatile` + double-checked locking for lazy initialization. While functionally correct in Java 5+, the `invalidateCache()` method sets both fields to `null` inside `synchronized`, but reads in `effectivePatterns()` and `compiledPatterns()` first check outside synchronized. There's a narrow window where `effectivePatterns` could be non-null but `compiledPatterns` is null after `invalidateCache()` is called, potentially causing a brief inconsistency (though likely benign since both get recomputed).

#### 4. 🟢 `InetAddress.getByName()` Network Call in Validation — Low Concern
The validation logic calls `InetAddress.getByName(host)` which performs a DNS lookup. In `isPrivateOrReservedHost()`, if DNS fails, it falls back to string pattern matching via `isPrivateIpv4Pattern()`. This is reasonable but means:
- Validation of URLs with unresolvable hostnames only checks string patterns (no IPv6 fallback)
- DNS resolution adds latency to every URL validation call
- DNS resolution could be used as a side-channel to confirm internal hostnames exist

#### 5. 🟢 Glob Pattern Edge Cases — Low Concern
- The `**` pattern translates to `(.*\.)?` which correctly matches "zero or more subdomains + dot", but the logic skips the next character if it's `.` (line `i++`). This is tested and appears correct.
- No maximum length check on patterns; extremely long patterns could cause regex backtracking, though the current pattern structure (no nested quantifiers) makes this unlikely.

#### 6. 🟢 No Admin API for Dynamic Allowlist Updates
Issue #1042 mentioned "via configuration/admin API" — the current implementation only supports configuration via `application.properties`. The REST endpoint is read-only (GET). Dynamic updates would require a restart. This is a reasonable V1 approach but should be noted.

---

### Alignment with Issue #1042 Requirements

| Requirement | Status | Notes |
|---|---|---|
| Configurable allowlist via `application.properties` | ✅ Done | `wanaku.toolset-repos.url-allowlist` property |
| Glob pattern support (`*.github.com`, etc.) | ✅ Done | `*`, `**`, `?` patterns supported |
| Only allowlisted hosts accepted when configured | ✅ Done | Tested thoroughly |
| Block localhost/private/reserved when no allowlist | ✅ Done | Enhanced with link-local, CGNAT, IPv6 |
| Redirect handling does not bypass allowlist | ❌ Not addressed | `URL.openStream()` follows redirects; no re-validation |
| Admin API for configuration | ⚠️ Partial | Read-only endpoint; no write/update API |

---

### Recommendations

1. **Critical:** Address redirect bypass — either disable automatic redirects in the HTTP connections used by `ToolsetRepoIndex.fromUrl()` and `ToolsetIndexHelper.loadToolsIndex()`, or implement a custom redirect handler that re-validates each redirect target against the allowlist. This could be done by using `HttpURLConnection` directly with `setInstanceFollowRedirects(false)` and manually following redirects with validation.

2. **Consider:** Add the redirect mitigation as a follow-up issue if not addressed in this PR, since it was an explicit requirement in #1042.

3. **Minor:** Add a note in the documentation that the allowlist does not currently protect against open redirects from allowed hosts, so users should be aware when allowlisting hosts that may offer redirect services.

orpiske:
## 🔍 Security Review: URL Allowlist for Toolset Repositories

Thanks for the solid work on this PR — the allowlist mechanism, glob pattern matching, expanded IP range checks, and test coverage are all great improvements. That said, a security-focused review uncovered several issues that should be addressed before merge. Ordered by severity:

---

### 🔴 CRITICAL — HTTP Redirect Bypass (SSRF)

**Files:** `ToolsetRepoIndex.java:67`, `ToolsetIndexHelper.java:39-40`

The validation in `ToolsetReposBean.validateUrl()` checks the *original* URL, but the actual HTTP connections (`URL.openStream()` and `ObjectMapper.readValue(URL, ...)`) follow HTTP redirects automatically. An attacker can register a repository pointing to an allowed host that 302-redirects to an internal service (e.g., `http://169.254.169.254/...` for cloud metadata).

**Flow:**
1. Attacker registers repo URL: `https://allowed-host.example.com/repo`
2. `validateUrl()` passes — host is allowlisted
3. `ToolsetRepoIndex.fromUrl()` → `URI.create(indexUrl).toURL().openStream()` follows redirect to `http://169.254.169.254/latest/meta-data/`
4. Allowlist is bypassed entirely

**Recommendation:** Replace `URL.openStream()` with `HttpURLConnection` (or `java.net.http.HttpClient`) configured with `setInstanceFollowRedirects(false)`. Manually handle redirects by extracting the `Location` header and re-validating the redirect target through `validateUrl()` before following it. Alternatively, use `HttpClient` with `HttpClient.Redirect.NEVER`.

---

### 🔴 HIGH — `toolsetUrl` Not Validated After Construction; Path Traversal Risk

**File:** `ToolsetReposBean.java:252-256`

```java
String baseUrl = resolveBaseUrl(ds);
validateUrl(baseUrl);
String toolsetUrl = ToolsetRepoIndex.toolsetUrl(baseUrl, toolsetName);
// toolsetUrl is used directly without re-validation
return ToolsetIndexHelper.loadToolsIndex(URI.create(toolsetUrl).toURL());
```

`toolsetName` comes from user input and is concatenated into the URL via `toolsetUrl()` (which does `baseUrl + "toolsets/" + toolsetName + ".json"`). A malicious `toolsetName` like `../../etc/passwd` or `../../../internal-api/endpoint` could construct a URL pointing outside the intended path, or even to a different host if combined with URL parsing quirks.

**Recommendation:**
- Sanitize `toolsetName` — reject names containing `/`, `\`, `..`, or any non-alphanumeric characters beyond `-` and `_`.
- Call `validateUrl(toolsetUrl)` *after* constructing the full toolset URL.

---

### 🟡 MEDIUM — DNS Rebinding / TOCTOU Gap

**File:** `ToolsetReposBean.java:307-332`

`isLoopbackHost()` and `isPrivateOrReservedHost()` resolve the hostname via `InetAddress.getByName()` at validation time. Between this DNS resolution and the actual HTTP connection (which resolves DNS again), an attacker-controlled DNS server could return a public IP first (passing validation) and then a private IP for the actual connection.

**Recommendation:** This is hard to fully mitigate at the application layer, but two steps help:
1. Resolve DNS once, then connect to the resolved IP directly (pin the IP).
2. Document this as a known limitation and recommend network-level controls (e.g., blocking outbound traffic to RFC 1918 ranges from the service).

---

### 🟡 MEDIUM — Allowlist Bypasses Private IP Checks Entirely

**File:** `ToolsetReposBean.java:289-301`

```java
if (urlAllowlistConfig.isAllowlistConfigured()) {
    if (!urlAllowlistConfig.isHostAllowed(host)) {
        throw new WanakuException("URL host '%s' is not in the allowlist...");
    }
    // ← No private IP check when allowlist is configured!
} else {
    if (isLoopbackHost(host)) { ... }
    if (isPrivateOrReservedHost(host)) { ... }
}
```

When the allowlist is configured, the private/reserved IP checks are skipped. If an admin accidentally allowlists a pattern like `*.internal.corp` or `*`, private IPs would be accepted.

**Recommendation:** Always run `isLoopbackHost()` and `isPrivateOrReservedHost()` checks *before* the allowlist check, or at minimum after it. The private IP check should be a hard gate regardless of allowlist configuration. Something like:

```java
if (isLoopbackHost(host)) {
    throw new WanakuException("URLs pointing to localhost are not allowed");
}
if (isPrivateOrReservedHost(host)) {
    throw new WanakuException("URLs pointing to private or reserved network ranges are not allowed");
}
if (urlAllowlistConfig.isAllowlistConfigured() && !urlAllowlistConfig.isHostAllowed(host)) {
    throw new WanakuException("URL host '%s' is not in the allowlist...");
}
```

---

### 🟡 MEDIUM — Double-Checked Locking Correctness

**File:** `UrlAllowlistConfig.java`

The class uses two separate `volatile` fields (`effectivePatterns` and `compiledPatterns`) with independent double-checked locking blocks. While each individual DCL block is technically correct for its own field, `compiledPatterns()` depends on `effectivePatterns()`. After `invalidateCache()` nulls both fields, a thread could see `compiledPatterns == null` but read a stale `effectivePatterns` from another thread's perspective if the fields are set in an interleaved manner.

**Recommendation:** Consider using a single immutable holder object for both fields, or use `@PostConstruct` initialization since the config is read-once-at-startup. For a `@ApplicationScoped` bean with config from MicroProfile Config, lazy initialization is likely unnecessary — eager init in `@PostConstruct` would be simpler and safer:

```java
@PostConstruct
void init() {
    this.effectivePatterns = parsePatterns(rawAllowlistPatterns);
    this.compiledPatterns = compilePatterns(this.effectivePatterns);
}
```

---

### 🟢 LOW — Minor Issues

1. **`isReservedIpv6` logic too broad:** Matching `bytes[0] == 0x00 && bytes[1] == 0x00` catches `::1` (loopback, already handled) and `::` (unspecified, already handled by `isAnyLocalAddress()`), but also catches valid addresses like `::ffff:x.x.x.x` (IPv4-mapped IPv6) and Teredo addresses. Consider being more precise or documenting the intent.

2. **`/allowlist` endpoint not authenticated:** The `getAllowlist` endpoint in `ToolsetReposResource` exposes the security configuration to any caller. If the rest of the API requires authentication, this endpoint should too.

3. **Missing OpenAPI annotations:** The new `/allowlist` endpoint lacks `@Operation`, `@Tag`, and `@APIResponse` annotations that the other endpoints presumably have.

4. **Exception handling fragility in `isPrivateOrReservedHost`:** `InetAddress.getByName()` failure falls back to string-pattern matching via `isPrivateIpv4Pattern()`. This is a reasonable fallback, but the DNS failure itself could mask issues. Consider logging a warning.

---

### 📋 Summary of Recommended Actions

| Priority | Issue | Action |
|----------|-------|--------|
| 🔴 Critical | Redirect bypass | Disable auto-redirects; re-validate redirect targets |
| 🔴 High | `toolsetUrl` unvalidated | Sanitize `toolsetName`; validate constructed URL |
| 🟡 Medium | DNS rebinding | Pin resolved IP; document limitation |
| 🟡 Medium | Allowlist skips private IP check | Always run private IP checks |
| 🟡 Medium | DCL correctness | Use eager init or single holder object |
| 🟢 Low | IPv6 check, auth, OpenAPI, logging | Address as cleanup |

The critical redirect bypass issue should be fixed before merge — it completely undermines the allowlist protection. The other items can be addressed incrementally, but the "always run private IP checks" fix is a quick win that significantly hardens the allowlist path.

Happy to help with any of these fixes or discuss alternatives!

orpiske:
## PR #1128 Review — URL allowlist for toolset repositories (SSRF mitigation)

This is a well-motivated security contribution. The architecture — blocklist-first then allowlist, redirect re-validation at each hop, toolset name validation — demonstrates good understanding of the SSRF threat model. Three critical findings should be addressed before merge.

---

### Critical

**C1. TOCTOU / DNS Rebinding bypass**
`validateUrl` resolves DNS via `InetAddress.getByName()` to check if the IP is private, but then `openValidatedStream` opens a *separate* connection that resolves DNS again. An attacker-controlled DNS with TTL=0 can return a public IP for the first query and `127.0.0.1` for the second. This completely defeats the SSRF protection. Fix: resolve DNS once, validate the resolved address, then connect to the resolved IP directly (setting the `Host` header to preserve virtual hosting).

**C2. `isReservedIpv6` is incorrect**
Checking `bytes[0]==0x00 && bytes[1]==0x00` only catches `::` and `::1` (already handled by `isLoopbackAddress()`/`isAnyLocalAddress()`). It misses: IPv4-mapped IPv6 addresses (`::ffff:127.0.0.1`, `::ffff:169.254.169.254`) and IPv6 Unique Local Addresses (`fc00::/7`). These are real SSRF bypass vectors.

**C3. `isLoopbackHost` fails open on `UnknownHostException`**
If DNS resolution fails (transient error or attacker-controlled DNS refusing the first query), the host passes all checks. Security validations should **fail closed** — reject unresolvable hosts rather than assume they're safe.

### Important

**I1. Resource leak in `openValidatedStream`** — On the success path, the `HttpURLConnection` is never `disconnect()`'ed. Wrap the returned `InputStream` in a `FilterInputStream` that calls `disconnect()` on close.

**I2. No `User-Agent` header** — Many CDNs (including GitHub) return `403` for requests without one. Add `connection.setRequestProperty("User-Agent", "Wanaku-Router/1.0")`.

**I3. Allowlist endpoint security** — `GET /api/v1/toolset-repos/allowlist` exposes security configuration. When deployed with Keycloak, this should likely require admin role.

**I4. Redundant catch block** — `catch (WanakuException e) { throw e; }` in `fetchToolset` is needed to avoid re-wrapping but deserves a brief comment explaining why.

### Suggestions

- **S1.** Glob pattern `*` alone produces `^[^.]*$` (matches single-label hosts only, not "everything") — may surprise users. Consider special-casing or documenting.
- **S2.** `**` glob is only meaningful as a prefix (`**.example.com`). Document this limitation.
- **S3.** Empty `@param`/`@return` tags on `ToolsetIndexHelper.loadToolsIndex(InputStream)`.
- **S4.** No test for IPv4-mapped IPv6 bypass vectors (`https://[::ffff:127.0.0.1]/repo`).
- **S5.** `URI.create(url)` is called twice in `openValidatedStream` — parse once and reuse.

### Verdict

The security intent and overall design are strong. The redirect re-validation and defense-in-depth ordering (blocklist before allowlist) are exactly right. However, the DNS rebinding TOCTOU (C1), incorrect IPv6 checks (C2), and fail-open behavior (C3) represent concrete bypass vectors that should be fixed before merge.

orpiske:
Please, can you rebase this one? I'd like to take another look at it. I'm hoping we can get it in 0.2.0 ...

orpiske (commented):
## PR #1128 Review: URL allowlist for toolset repositories to mitigate SSRF risks

Solid security enhancement. The PR replaces naive string-based IP checks with proper `InetAddress` resolution, adds DNS-rebinding (TOCTOU) protection by connecting to the resolved IP, introduces glob-based URL allowlisting, validates toolset names against path traversal, and covers it all with good tests and documentation.

---

### Confirmed Issues

**1. Dead code: `isPrivateIpv4Pattern` (medium)**
`ToolsetReposBean.isPrivateIpv4Pattern()` is defined but never called anywhere. The PR switched from string-based host matching to `InetAddress`-based resolution, making this method unnecessary. Tests exist for it (`ToolsetReposBeanUrlValidationTest:206-214`) which exercise dead code.

**2. `ToolsetRepoIndex.fromUrl()` remains as an unprotected fetch path (medium)**
The PR replaces `ToolsetRepoIndex.fromUrl(url)` with `fetchToolsetRepoIndex(url)` which validates URLs and controls the connection. However, `fromUrl()` at `core/core-services-api/.../ToolsetRepoIndex.java:65-72` still exists and performs `URI.create(indexUrl).toURL().openStream()` with no SSRF validation. A future caller could use it and bypass all the protections this PR adds. Consider deprecating it or adding a Javadoc warning.

**3. Discarded `ResolvedUrl` in `browse()` and `fetchToolset()` (low)**
In both methods, the outer `validateUrl()` calls discard the returned `ResolvedUrl`:

```java
// browse()
validateUrl(url);                                     // ResolvedUrl discarded
ToolsetRepoIndex index = fetchToolsetRepoIndex(url);   // re-resolves DNS internally

// fetchToolset()
validateUrl(baseUrl);    // ResolvedUrl discarded
...
validateUrl(toolsetUrl); // ResolvedUrl discarded
```

The DNS-rebinding protection only works when the resolved IP is used for the actual connection, which happens inside `openValidatedStream`. The outer calls serve as fail-fast gates (good for UX), but the discarded returns may confuse a reader who sees the Javadoc claiming TOCTOU protection. Not a security bug — the inner call in `openValidatedStream` provides the actual protection — but slightly misleading.

**4. Reflection-based test setup is fragile (low)**
`ToolsetReposBeanRedirectSecurityTest` uses `Field.setAccessible(true)` to inject `dataStoreRepository` but never injects `urlAllowlistConfig`. This works because the anonymous subclass overrides `validateUrl()`, but if that override is ever removed, the test will NPE on the uninjected field rather than giving a clear diagnostic. The other test class (`ToolsetReposBeanUrlValidationTest`) uses `@InjectMocks` consistently and is more robust.

---

### Questions / Assumptions

1. **Should `ToolsetRepoIndex.fromUrl()` be deprecated?** Since this PR is specifically about SSRF hardening, leaving an unprotected fetch method in the public API seems like a gap worth addressing.

2. **`HttpURLConnection` vs Quarkus HTTP client** — Using `java.net.HttpURLConnection` directly is a deliberate choice for controlling connection behavior (connect to resolved IP, disable auto-redirect). This is reasonable for security-sensitive code, just noting it as an intentional trade-off.

3. **Glob pattern `*` semantics** — Documented correctly: `wanaku.toolset-repos.url-allowlist=*` combined with the IP blocklist means "all public hosts." Worth confirming this is the intended behavior since it makes the allowlist a noop.

---

### What's Done Well

- DNS-rebinding protection via `ResolvedUrl` + connecting to the resolved IP is a strong SSRF mitigation pattern
- Comprehensive IP range coverage: loopback, private, link-local, CGNAT, IPv4-mapped IPv6, unique local
- Redirect re-validation prevents redirect-based SSRF bypasses
- Toolset name validation (`^[A-Za-z0-9_-]+$`) prevents path traversal
- Test coverage is thorough: 3 new test classes, parameterized tests, redirect security test with an embedded HTTP server
- Documentation in `configurations.md` clearly explains the pattern semantics with examples

---

> Note: This review covers project conventions and correctness. It does not replace specialized tools like CodeRabbit, Sourcery, or static analysis.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/toolsetrepos/ToolsetReposBean.java
```diff
@@ -326,6 +466,93 @@ static String toRawContentUrl(String url, String branch) {
...
+        }
+        int port = uri.getPort();
+
+        URL connectUrl = new URL(
+                uri.getScheme(),
+                ipHost,
+                port == -1 ? -1 : port,
+                uri.getPath()
+                        + (uri.getQuery() != null ? "?" + uri.getQuery() : "")
+                        + (uri.getFragment() != null ? "#" + uri.getFragment() : ""));
```
orpiske: **HTTPS/TLS issue:** This constructs a URL with the resolved IP as the host (e.g. `https://93.184.216.34/path`). For HTTPS, `HttpsURLConnection` will use this IP for SNI and certificate verification, which will fail because the server certificate is issued for the hostname (e.g. `*.githubusercontent.com`), not the IP.

Setting the `Host` header (line 513) only affects the HTTP layer, not the TLS handshake.

Suggested approach: perform DNS validation (resolve + check private ranges) but connect using the original hostname. This still blocks SSRF via private IPs while keeping TLS working. Full DNS-rebinding protection with IP pinning would require a custom `SSLSocketFactory` with hostname-based SNI.

orpiske: **Deprecated API:** `new URL(String, String, int, String)` is deprecated since Java 20. Consider constructing via `URI` and calling `toURL()`.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/toolsetrepos/ToolsetReposResource.java
```diff
@@ -81,4 +84,10 @@ public WanakuResponse<List<ToolReference>> fetchToolset(
         LOG.debugf("REST: Fetching toolset '%s' from repository: %s", toolsetName, name);
         return new WanakuResponse<>(toolsetReposBean.fetchToolset(name, toolsetName));
     }
+
+    @Path("/allowlist")
+    @GET
+    public WanakuResponse<List<String>> getAllowlist() {
```
orpiske, line 90: Minor API design note: since the existing endpoints use `{name}` as a path parameter (e.g. `PUT /api/v1/toolset-repos/{name}`), a repository named `"allowlist"` would collide with this path. A pattern like `/api/v1/toolset-repos/-/allowlist` avoids the ambiguity.

orpiske (changes requested):
Thanks for this thorough SSRF mitigation work, @matheusandre1! The coverage across private IP ranges, IPv6, CGNAT, redirect re-validation, and path traversal prevention is impressive, and the test suite is excellent.

I've flagged one blocking issue around HTTPS/TLS and a few smaller items below.

---

**Confirmed issues:**

1. **HTTPS connections will fail due to SNI mismatch (High)** — `openValidatedStream` connects to the resolved IP address directly (for DNS-rebinding TOCTOU protection) by constructing a URL like `https://<IP>:<port>/path` and setting the `Host` header. However, for HTTPS, Java's `HttpsURLConnection` uses the URL's host for TLS SNI negotiation and certificate verification. Since the URL host is an IP address (not the original hostname), the server's certificate (issued for e.g. `*.githubusercontent.com`) won't match, causing TLS handshake failures. This affects all `browse()` and `fetchToolset()` operations since `toRawContentUrl()` produces `https://raw.githubusercontent.com/...` URLs. The simplest fix: validate the URL (DNS resolution + private IP check) separately, then connect using the original hostname — this still prevents the most common SSRF vectors while avoiding the TLS issue.

2. **`new URL()` constructor deprecated since Java 20 (Low)** — `openValidatedStream` uses `new URL(scheme, host, port, file)` which is deprecated in Java 20+. The project targets Java 21.

**Questions / suggestions:**

3. **API path collision** — The new `GET /api/v1/toolset-repos/allowlist` uses the same namespace as `{name}` in existing endpoints. A repository named "allowlist" would shadow the endpoint (or vice versa). Consider a path like `/api/v1/toolset-repos/-/allowlist` to avoid collision.

4. **`add()`/`update()` consistency** — These call `validateUrl()` directly while `browse()`/`fetchToolset()` call `validateUrlFailFast()`. Functionally identical, but reads inconsistently.

---

orpiske:
Hey @matheusandre1, can you rebase this one more time, please? I think we're getting close to merging this one ...

matheusandre1:
> Hey @matheusandre1, can you rebase this one more time, please? I think we're getting close to merging this one ...

I think we can close this pull request; there was a change that improved and fixed this...

orpiske:
> > Hey @matheusandre1, can you rebase this one more time, please? I think we're getting close to merging this one ...
> 
> I think we can close this pull request; there was a change that improved and fixed this...

Ahhh, ok. Let's close it. Thanks!

PR #1170, opened 2026-05-20
[UI] Optional prompt modal fields in own tabs

rnetuka:
I've moved the two optional fields in Prompt Modal into own tabs. They are marked as optional so the user should be able to submit the dialog witout even looking at them.

## Summary by Sourcery

Extract the prompt creation modal into a dedicated component and reorganize its fields into tabbed sections to separate core and optional inputs.

New Features:
- Introduce a reusable PromptModal component for creating prompts with tabbed navigation for different input sections.

Enhancements:
- Move optional prompt arguments and tool reference fields into their own tabs to streamline the primary prompt creation flow.
- Add a scrollable wrapper for the prompt modal tab panels to maintain a consistent modal height.

### DIFF
--- a/apps/ui/admin/src/App.scss
+++ b/apps/ui/admin/src/App.scss
@@ -28,3 +28,8 @@ li {
   height: 460px;
   overflow-y: auto;
 }
+
+#prompts-tab-panel-wrapper {
+  height: 460px;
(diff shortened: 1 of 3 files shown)


orpiske (approved):
This looks a lot better. Thanks!

PR #1175, opened 2026-05-22
[#1173] LLM model field cannot be edited in Firefox

### LINKED ISSUE #1173: LLM model field cannot be edited in Firefox
## Description

The LLM model field in the chat/prompt dialog cannot be edited when using Firefox. The field appears to be read-only and also renders as a password field.

## Steps to Reproduce

1. Open the admin UI in Firefox
2. Navigate to the chat/prompt dialog
3. Try to edit the LLM model field

## Expected Behavior

The LLM model field should be a regular editable text input, allowing the user to type or change the model name.

## Actual Behavior

The field appears read-only (cannot be clicked or typed into) and renders as a password field.

## Additional Context

- Browser: Firefox
- Other browsers: not tested

rnetuka:
Issue: https://github.com/wanaku-ai/wanaku/issues/1173

I've removed the _useRef_ hook, since on second look, the same effect could be achieved with more simple approach. Also, it seems React has problem with ComboBox rendering. I had to add setTimeout in the handler in order not to call the value change immidiately. 

I hope it fixes the issue!

## Summary by Sourcery

Simplify LLM model selection wiring in the LLM chat admin page and fix the model ComboBox behavior, particularly in Firefox.

Bug Fixes:
- Resolve LLM model ComboBox value update issues by deferring change handling to avoid React/ComboBox rendering glitches.

Enhancements:
- Replace ref-based imperative updates with a config-driven LLM model ComboBox that derives suggestions from the current base URL.
- Streamline LLM setup configuration flow so model and base URL changes are propagated through a single config object.

### DIFF
--- a/apps/ui/admin/src/Pages/LLMChat/LLMChatPage.tsx
+++ b/apps/ui/admin/src/Pages/LLMChat/LLMChatPage.tsx
@@ -16,7 +16,12 @@ export const LLMChatPage: React.FC = () => {
       <h1 className="title">LLM Chat for testing</h1>
       <Grid fullWidth>
         <Column lg={4}>
-          <LLMSetup config={config} onChange={setConfig} />
+          <LLMSetup
+            config={config}
+            onChange={(config: LlmConfig) => {
+              setConfig({ ...config })}
+            }
+          />
           <LLMTools
             selectedTools={config.tools}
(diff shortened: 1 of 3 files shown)


orpiske:
Just tested it here ... working very well now. Thanks!

PR #1188, opened 2026-05-22
feat: PlaceholderResolver for dynamic placeholder resolution

### LINKED ISSUE #1143: Support version placeholders in service catalogs
## Description

Currently, in the service catalog, Camel (and other dependencies) versions are specified directly. This makes maintenance very hard as versions must be updated in every catalog individually.

## Proposed Solution

Use version markers/placeholders such as `${camel.version}` and others in service catalog definitions, and then pre/post-process the catalog to resolve them with the actual version in use.

## Requirements

- Support placeholders like `${camel.version}` for dependency versions in service catalogs
- Pre/post-process catalogs to resolve placeholders to actual versions at the appropriate stage
- Continue supporting hard-coded versions so edge cases relying on specific versions still work
- Placeholders should be optional — catalogs with explicit versions remain valid

## Additional Context

This aligns with reducing maintenance burden as the number of service catalogs grows.

matheusandre1:
Closes: #1143

## Summary by Sourcery

Introduce version placeholder resolution for service catalog dependency files via a shared versions properties file and integrate it across CLI commands, backend deployment, and templates.

New Features:
- Add PlaceholderResolver utility for resolving ${placeholder} variables from properties-based bindings.
- Extend ServiceCatalogIndex to support an optional catalog.versions file and resolve placeholders in dependency files within ZIP archives.
- Add --resolve-versions option to service packaging and deployment CLI commands to substitute dependency placeholders before creating ZIPs.
- Initialize new service catalogs with a versions.properties skeleton and wire existing service templates to use a shared versions.properties file for dependency versions.

Enhancements:
- Validate referenced catalog.versions files in ServiceCatalogIndex and expose helper methods to query their presence and path.
- Resolve dependency placeholders in router backend catalog/template deployment flows when a versions file is present.
- Refactor and extend ServiceCatalogIndex tests to cover versions file handling and placeholder resolution behavior.

Tests:
- Add unit tests for PlaceholderResolver covering resolution, placeholder detection, and bindings loading.
- Add ServiceCatalogIndex tests for versions file discovery, validation, and end-to-end placeholder resolution in dependency files.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/service/ServiceDeploy.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/service/ServiceDeploy.java
@@ -7,6 +7,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.Properties;
 import java.util.zip.ZipEntry;
@@ -16,6 +17,7 @@ import ai.wanaku.capabilities.sdk.api.types.DataStore;
 import ai.wanaku.capabilities.sdk.api.types.WanakuResponse;
 import ai.wanaku.cli.main.commands.BaseCommand;
 import ai.wanaku.cli.main.support.WanakuPrinter;
+import ai.wanaku.core.services.api.ServiceCatalogIndex;
 import ai.wanaku.core.services.api.ServiceCatalogService;
 import ai.wanaku.core.services.api.ServiceTemplateService;
 import picocli.CommandLine;
@@ -44,6 +46,12 @@ public class ServiceDeploy extends BaseCommand {
             description = "Deploy as a service template instead of a catalog",
             defaultValue = "false")
     private boolean template;
+
+    @CommandLine.Option(
+            names = {"--resolve-versions"},
+            description = "Resolve ${placeholder} versions in dependency files before deploying",
+            defaultValue = "true")
+    private boolean resolveVersions;
 
     @Override
     public Integer doCall(Terminal terminal, WanakuPrinter printer) throws Exception {
@@ -113,6 +121,27 @@ public class ServiceDeploy extends BaseCommand {
         } catch (IOException e) {
             printer.printErrorMessage(String.format("Failed to create ZIP archive: %s%n", e.getMessage()));
             return EXIT_ERROR;
+        }
+
+        // Resolve version placeholders in-memory, never on disk.
+        if (resolveVersions) {
+            String versionsPath = props.getProperty("catalog.versions");
+            if (versionsPath != null) {
+                File versionsFile = new File(baseDir, versionsPath);
+                if (versionsFile.exists()) {
+                    try {
+                        ServiceCatalogIndex index = ServiceCatalogIndex.fromZipBytes(zipBytes);
+                        byte[] resolvedZip = index.resolvePlaceholders(zipBytes);
+                        if (!Arrays.equals(resolvedZip, zipBytes)) {
+                            zipBytes = resolvedZip;
+                        }
+                    } catch (Exception e) {
+                        printer.printErrorMessage(
+                                String.format("Failed to resolve version placeholders: %s%n", e.getMessage()));
+                        return EXIT_ERROR;
+                    }
+                }
+            }
         }
 
         // Base64-encode the ZIP
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/service/ServiceInit.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/service/ServiceInit.java
@@ -89,6 +89,7 @@ public class ServiceInit extends BaseCommand {
         props.setProperty("catalog.name", name);
         props.setProperty("catalog.description", "Service catalog for " + name);
         props.setProperty("catalog.services", services);
+        props.setProperty("catalog.versions", "versions.properties");
 
         for (String system : systems) {
             String systemName = system.trim();
@@ -100,7 +101,6 @@ public class ServiceInit extends BaseCommand {
             props.setProperty(
                     "catalog.dependencies." + systemName, systemName + "/" + systemName + ".dependencies.txt");
 
-            // If --template flag is set, add catalog.properties entry
             if (template) {
                 props.setProperty("catalog.properties." + systemName, systemName + "/service.properties");
             }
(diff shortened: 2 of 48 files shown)


apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/service/ServiceDeploy.java
```diff
@@ -177,4 +206,30 @@ private void addDirectoryToZip(ZipOutputStream zos, File dir, String prefix) thr
...
+            if (!depsFile.exists()) {
+                continue;
+            }
+
+            String content = Files.readString(depsFile.toPath(), StandardCharsets.UTF_8);
+            String resolved = PlaceholderResolver.resolve(content, bindings);
+            if (!resolved.equals(content)) {
+                Files.writeString(depsFile.toPath(), resolved, StandardCharsets.UTF_8);
+            }
+        }
```
orpiske: **Bug:** `Files.writeString(depsFile.toPath(), resolved)` permanently overwrites the user's source file on disk. After running `deploy`, all `${placeholder}` values in the original dependency files are replaced with hard-coded versions. Running the command again finds nothing to resolve.

The resolution should happen in-memory (e.g., resolve content before writing to the `ZipOutputStream`) rather than mutating files on disk.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/service/ServicePackage.java
```diff
@@ -143,4 +174,30 @@ private void addDirectoryToZip(ZipOutputStream zos, File dir, String prefix) thr
...
+            if (!depsFile.exists()) {
+                continue;
+            }
+
+            String content = Files.readString(depsFile.toPath(), StandardCharsets.UTF_8);
+            String resolved = PlaceholderResolver.resolve(content, bindings);
+            if (!resolved.equals(content)) {
+                Files.writeString(depsFile.toPath(), resolved, StandardCharsets.UTF_8);
+            }
+        }
```
orpiske: Same issue as in `ServiceDeploy` — this writes resolved content back to the original source file, permanently removing placeholders.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/servicecatalog/ServiceCatalogBean.java
```diff
@@ -105,7 +106,16 @@ public DataStore deploy(DataStore dataStore) throws WanakuException {
         }
 
         // Validate ZIP structure by parsing the index
-        ServiceCatalogIndex.fromBase64(dataStore.getData());
+        ServiceCatalogIndex index = ServiceCatalogIndex.fromBase64(dataStore.getData());
+
+        // Resolve version placeholders in dependency files if a versions file is present
+        if (index.hasVersionsFile()) {
+            byte[] zipBytes = Base64.getDecoder().decode(dataStore.getData());
```
orpiske, line 114: `resolvedZip != zipBytes` uses reference equality on byte arrays. This works because `resolvePlaceholders` returns the same reference when no changes occur, but it's fragile. If the implementation is ever refactored to always produce a new array, this check silently breaks. Consider returning a wrapper or boolean flag instead.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/service/ServiceDeploy.java
```diff
@@ -177,4 +206,30 @@ private void addDirectoryToZip(ZipOutputStream zos, File dir, String prefix) thr
             }
         }
     }
```
orpiske, line 208: This method is identical to `ServicePackage.resolveDependencyPlaceholders`. Consider extracting it to `PlaceholderResolver` or a shared base class to avoid drift.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/service/ServiceInit.java
```diff
@@ -110,6 +110,15 @@ private void createIndexProperties(File rootDir, String[] systems) throws IOExce
...
+
+        // Create versions.properties skeleton
+        File versionsFile = new File(rootDir, "versions.properties");
+        try (PrintWriter pw = new PrintWriter(new FileWriter(versionsFile))) {
+            pw.println("# Version placeholders for dependency files");
+            pw.println("# Reference these in *.dependencies.txt as ${name}");
+            pw.println("# Example: camel.version=4.18.2");
+            pw.println("camel.version=4.18.2");
+        }
     }
```
orpiske, line 123: Hardcoding `camel.version=4.18.2` in the skeleton will go stale as Camel releases new versions. Consider using a placeholder comment or deriving it from the project.

core/core-services-api/src/main/java/ai/wanaku/core/services/api/ServiceCatalogIndex.java
```diff
@@ -28,9 +31,17 @@
  * </ul>
  * Optional properties:
  * <ul>
- *   <li>{@code catalog.icon} - display icon</li>
- *   <li>{@code catalog.dependencies.<system>} - relative path to dependencies file</li>
+ * <li>{@code catalog.icon} - display icon</li>
```
orpiske: Nit: This Javadoc indentation change (`*   <li>` → `* <li>`) is unrelated to the feature. Minor, but adds noise to the diff.

orpiske (changes requested):
Thanks for tackling #1143, @matheusandre1 — this is a solid feature and the `PlaceholderResolver` utility is clean and well-tested. I have a few concerns, the most important being a bug where the CLI commands mutate source files on disk.

> **Note:** This review covers project conventions and code-level concerns. It does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud.

### Confirmed Issues

**1. CLI commands mutate source files on disk (Bug)**

`ServiceDeploy` and `ServicePackage` call `Files.writeString(depsFile.toPath(), resolved)` which **overwrites the user's original dependency files** with resolved versions *before* creating the ZIP. After running `wanaku service deploy` or `wanaku service package`, the `${camel.version}` placeholders in the source files are permanently replaced with hard-coded versions. Running the command a second time finds nothing to resolve. The resolution should happen in-memory during ZIP creation, not on the filesystem.

**2. Reference equality (`!=`) on byte arrays**

`ServiceCatalogBean.java` and `ServiceTemplateBean.java` use `if (resolvedZip != zipBytes)` which relies on identity comparison. This works today because `resolvePlaceholders` returns the same reference when unchanged, but it's fragile — if the implementation changes to always copy, this silently breaks. Consider using a boolean return or wrapper result.

**3. Code duplication**

`resolveDependencyPlaceholders` is copy-pasted between `ServiceDeploy` and `ServicePackage`. Extract it to a shared utility to prevent drift.

**4. Missing `versions.properties` per template directory**

All template `index.properties` files reference `catalog.versions=versions.properties` as a relative path. The new shared file sits at `services/service-templates/src/main/services/versions.properties` — one level above each template's root directory. How does this file end up inside each template's ZIP archive?

**5. Hardcoded version in `ServiceInit.java`**

`pw.println("camel.version=4.18.2")` bakes a specific version into every new skeleton. This will go stale.

### Questions

1. `PLACEHOLDER_PATTERN` uses `[\\w.]+` which excludes hyphens. Names like `${camel-version}` won't match — is this intentional?
2. Both CLI commands default `--resolve-versions=true`, and the router backend also resolves on deploy. Is the double resolution intentional, or should only one side resolve?

orpiske:
Also, this is likely going to 0.2.0. In the mean time, can you please rebase the code?

orpiske:
Hey @matheusandre1 thanks for looking at this one. This time, however, I decided to take a different approach directly on the SDK ... After reviewing what I needed and what our users would feel more comfortable with, I decided for an approach that is similar to what Camel does, so I am closing this one without merge.

PR #1212, opened 2026-05-26
Add a sample user to wanaku realm in keycloak config

claudio4j:
## Summary by Sourcery

Document and wire up a sample user in the Wanaku Keycloak realm configuration for local development.

Enhancements:
- Annotate the Keycloak service in docker-compose with a comment about the bundled sample user credentials.

Documentation:
- Add documentation for the sample Keycloak user credentials in the contributing guide.

### DIFF
--- a/deploy/auth/wanaku-config.json
+++ b/deploy/auth/wanaku-config.json
@@ -486,6 +486,34 @@
       ],
       "notBefore": 0,
       "groups": []
+    },
+    {
+      "id": "c8f9e2a1-3b4d-4c5e-8f7a-9d6e5b4c3a2b",
+      "username": "wuser",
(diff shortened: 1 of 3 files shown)


orpiske:
I'm a bit concerned with one thing: typically, this Keycloak config is the same one used as a basis for production deployments ... so this means we are now creating one standard user that admins may miss / forget to remove. 

Given that a user can be created using this:

```
wanaku admin users add --admin-username admin --admin-password <pass>   --username wktest --password secretpass --email wktest@example.com
```

Can we adjust the process a bit, so that it runs this command instead of modifying the reference config?

claudio4j:
I will close this, since there is the "user management" section in usage.md that already covers this.

PR #1213, opened 2026-05-26
Fix NPE in local runner when optional CLI options are null

orpiske:
## Summary
- `wanaku start local` crashes when `--capabilities-client-secret` is not provided because `withServiceOption()` stored null values in the environment map, and `ProcessBuilder.environment().putAll()` rejects nulls with a `NullPointerException`
- After fixing the NPE, the router backend still fails to start in noauth mode because `AuthConfigSource` disabled discovery on named OIDC tenants (`mcp`, `ns-*`) without disabling the tenants themselves, causing `Either 'jwks-path' or 'introspection-path' must be set`
- Now `withServiceOption()` skips null values, and `AuthConfigSource` sets `tenant-enabled=false` on all named tenants in noauth mode

## Test plan
- [ ] Run `wanaku start local` without `--capabilities-client-secret` and verify it starts successfully
- [ ] Run `wanaku start local --capabilities-client-secret <secret>` and verify the secret is passed through
- [ ] Verify `AuthConfigSourceTest` passes (71 tests pass)

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/runner/local/LocalRunner.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/runner/local/LocalRunner.java
@@ -28,7 +28,9 @@ public class LocalRunner {
         public LocalRunnerEnvironment() {}
 
         public LocalRunnerEnvironment withServiceOption(String key, String value) {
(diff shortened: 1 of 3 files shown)


orpiske:
This is not what we need to solve the start local problem.

PR #1222, opened 2026-05-27
fix: poll local router readiness before starting services

### LINKED ISSUE #1221: Replace fixed timeout in 'wanaku start local' with readiness/liveness polling
## Summary

`wanaku start local` currently waits a fixed 5 seconds for the router to become ready before proceeding. Since we now have readiness/liveness health check endpoints, we should poll those instead for more accurate and reliable startup detection.

## Current behavior

A hardcoded 5-second sleep is used after launching the router process, regardless of whether the router is actually ready to serve requests.

## Proposed behavior

Replace the fixed timeout with a polling loop that checks the router's readiness endpoint (e.g., `/q/health/ready`). The loop should:

- Poll at a short interval (e.g., every 500ms)
- Have a configurable maximum timeout (with a reasonable default)
- Fail fast with a clear error message if the timeout is exceeded

## Motivation

- The fixed timeout is fragile: too short on slow machines, unnecessarily slow on fast ones
- Readiness/liveness endpoints already exist and provide the ground truth for router availability

iccccccccccccc:
Fixes #1221

## Changes
- replace the fixed `wanaku start local` router sleep with polling against `/q/health/ready`
- keep `wanaku.cli.router-start-wait-secs` as the maximum readiness wait
- start capability services as soon as the router reports ready, and fail with a clear timeout if it never does
- add focused tests for eventual readiness and timeout behavior

## Verification
- `.\mvnw.cmd -pl apps/wanaku-cli -am '-Dtest=ai.wanaku.cli.runner.local.LocalRunnerTest' '-Dsurefire.failIfNoSpecifiedTests=false' test`
- `.\mvnw.cmd -pl apps/wanaku-cli -am -DskipTests compile`
- `git diff --check`

## Summary by Sourcery

Poll the local router readiness endpoint before starting services instead of waiting a fixed delay, enforcing a configurable timeout and covering the behavior with tests.

Bug Fixes:
- Ensure local capability services start only after the router actually reports readiness, avoiding race conditions from a fixed startup sleep.
- Fail fast with a clear error when the local router never becomes ready within the configured wait time.

Enhancements:
- Introduce configurable HTTP-based readiness polling for the local router with a sensible default health endpoint and poll interval.
- Document the router startup wait configuration as a maximum readiness wait rather than a fixed delay.

Tests:
- Add unit tests verifying that readiness polling succeeds once the router becomes healthy and fails when it never does.

### DIFF
--- a/apps/wanaku-cli/src/test/java/ai/wanaku/cli/runner/local/LocalRunnerTest.java
+++ b/apps/wanaku-cli/src/test/java/ai/wanaku/cli/runner/local/LocalRunnerTest.java
@@ -6,6 +6,7 @@ import java.net.URI;
 import java.net.http.HttpClient;
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.LockSupport;
 import ai.wanaku.cli.main.support.WanakuCliConfig;
 import com.sun.net.httpserver.HttpServer;
 
@@ -13,6 +14,7 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
 
 class LocalRunnerTest {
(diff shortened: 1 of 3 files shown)


orpiske:
Thanks for this! It's on my TODO list to review this and other PRs after I complete our post release checklist.

iccccccccccccc:
Thanks, no rush. I just rebased this on current `main` so it should no longer be blocked by the release commits.

I also kept the newer local profile startup option from `main`, and folded in the Sourcery notes around interruption handling and per-request timeout bounds.

Reran locally:
- `./mvnw.cmd --% -Dmaven.repo.local=F:/work/.codex-tools/m2 -pl apps/wanaku-cli -am -Dtest=LocalRunnerTest -Dsurefire.failIfNoSpecifiedTests=false test`

apps/wanaku-cli/src/test/java/ai/wanaku/cli/runner/local/LocalRunnerTest.java
```diff
@@ -0,0 +1,98 @@
...
+        server.start();
+        URI readinessUri = URI.create("http://localhost:%d/q/health/ready"
+                .formatted(server.getAddress().getPort()));
+
+        LocalRunner runner = new LocalRunner(
+                mock(WanakuCliConfig.class),
+                new LocalRunner.LocalRunnerEnvironment(),
+                HttpClient.newHttpClient(),
+                readinessUri);
+
```
orpiske, line 90: Nit: timing-sensitive assertion — the 100ms timeout vs 500ms server delay vs 1s threshold gives good margin, but keep an eye on this test if CI flakiness appears on loaded runners.

orpiske (approved):
Solid contribution, @iccccccccccccc! The polling implementation is well thought out — proper deadline tracking, defensive timing, injectable dependencies for testing, and real HTTP servers in tests rather than mocks. Well done.

One minor finding below, nothing blocking.

**Behavioral note:** `run()` now throws `IOException` when the router times out, which propagates through `start()`. Previously the fixed sleep always succeeded and services were started regardless. This is the correct behavior (fail fast), just worth being aware of for callers.

This review covers project conventions and structural consistency. It does not replace specialized tools like CodeRabbit, Sourcery, or SonarCloud.

PR #1251, opened 2026-06-02
LLM chat with LangChain4j

rnetuka:
I've moved the LLM chat logic from UI to backend and implemented in in LangChain4j. I believe this brings more control and options in the future.

What is implemented:
- the chat is completely handled by _LangChain4j_. Only final answer is returned to the UI.
- LLM selection now displays names instead of URLs
- LLM allowlist is now an intersection between supported and allowed LLMs
- bugfix included - `DefaultMcpBridge` was storing already closed MCPClient that wasn't able to execute any tools
- The Stop button was removed because the UI method now awaits for the whole answer, so there is no purpose for it anymore.

What is not implemented:
- it's still possible to call public MCP tools only. I hope this opens the way to integration with Keycloak.

Any comments are welcome. 

@orpiske I have tested it with a tool call and with chat memory. Everythiong seems ok for me, but please test it before merging as these are big changes and I might have forgotten something or introduced a new bug. Thank you!

## Summary by Sourcery

Move LLM chat handling from the UI to the backend using LangChain4j with support for multiple providers and MCP tools.

New Features:
- Add backend LangChain4j-based chat endpoint that orchestrates LLM calls, tool usage via MCP, and short-term chat memory.
- Expose APIs to list allowed LLM providers and their models for dynamic UI configuration.

Bug Fixes:
- Fix MCP tool execution by creating a fresh MCP client per call instead of reusing a closed client instance.

Enhancements:
- Update UI LLM chat flow to call the backend chat endpoint and render only the final assistant response, removing client-side tool orchestration and streaming control.
- Replace URL-based LLM selection with named providers and dynamically populated model lists based on backend-reported capabilities.
- Introduce configuration mapping and helpers to pass custom LLM parameters such as max tokens, temperature, and tool choice from the UI to the backend.

Build:
- Add LangChain4j dependencies for Mistral, OpenAI, Anthropic, Google Gemini, and Ollama in the backend module.

### DIFF
--- a/apps/ui/admin/src/Pages/LLMChat/LLMChatArea.tsx
+++ b/apps/ui/admin/src/Pages/LLMChat/LLMChatArea.tsx
@@ -7,10 +7,10 @@ import {
   TextArea,
   Tile
 } from "@carbon/react"
-import {Send, Stop} from "@carbon/icons-react"
+import {Send} from "@carbon/icons-react"
 import {LlmConfig} from "./config"
-import {McpClient} from "./mcp"
 import {LLMChatMessage} from "./LLMChatMessage"
+import {getUrl} from "../../custom-fetch"
 
 
 interface ChatMessage {
@@ -36,9 +36,7 @@ export const LLMChatArea: React.FC<LLMChatAreaProps> = ({ config }) => {
   const [systemPrompt, setSystemPrompt] = useState("You are an assistant that can use tools.")
   const [userPrompt, setUserPrompt] = useState("")
   const [displayMessages, setDisplayMessages] = useState<DisplayMessage[]>([])
-  const [isRunning, setIsRunning] = useState(false)
   
-  const stopRunning = useRef(false)
   const chatMessages = useRef<ChatMessage[]>([])
   
   function clear() {
@@ -50,42 +48,8 @@ export const LLMChatArea: React.FC<LLMChatAreaProps> = ({ config }) => {
     
     const messages: DisplayMessage[] = [...displayMessages]
     
-    const tools = config.tools.map(tool => {
-      return {
-        type: "function",
-        function: {
-          name: tool.name,
-          description: tool.description || "No description provided.",
-          parameters: tool.inputSchema || {}
-        }
-      }
-    })
-    
-    async function chat(): Promise<Response> {
-      return fetch(window.location.origin + "/api/v1/chat/completions", {
-        method: "POST",
-        headers: {
-          "Content-Type": "application/json"
-        },
-        body: JSON.stringify({
-          baseUrl: config.baseUrl,
-          apiKey: config.apiKey,
-          chatParams: {
-            model: config.llmModel,
-            messages: chatMessages.current,
-            tools,
-            ...(config.extraLlmParams ? JSON.parse(config.extraLlmParams) : {})
-          }
-        })
-      })
-    }
-    
     function isFirstMessage() {
       return chatMessages.current.length === 0
-    }
-    
-    function isToolAvailable(tool: string): boolean {
-      return config.tools.map(tool => tool.name).includes(tool)
     }
     
     function displayMessage(message: DisplayMessage) {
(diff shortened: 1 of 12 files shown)


orpiske:
Cool. Thanks @rnetuka ... added to my review TODO.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java
```diff
@@ -3,88 +3,185 @@
...
-            if (apiKey != null && !apiKey.isEmpty()) {
-                requestBuilder.header("Authorization", "Bearer " + apiKey);
-            }
-            HttpRequest request = requestBuilder.build();
-            HttpResponse<String> response = client.send(request, BodyHandlers.ofString());
 
-            if (response.statusCode() != OK.getStatusCode()) {
-                LOG.errorf("HTTP Response Error: %s", response.body());
+        McpTransport transport = StreamableHttpMcpTransport.builder()
+                .url("http://localhost:8080/public/mcp")
```
orpiske, line 117: This hardcoded URL will only work in local development. In containerized or multi-host deployments, `localhost:8080` won't resolve to the router. This should be a configurable property (e.g., `wanaku.chat.mcp-url`), or derived from the router's own base URL.

```suggestion
                .url(mcpUrl)
```

(With a corresponding `@ConfigProperty(name = "wanaku.chat.mcp-url", defaultValue = "http://localhost:8080/public/mcp") String mcpUrl;` field.)

apps/ui/admin/src/Pages/LLMChat/LLMChatArea.tsx
```diff
@@ -50,127 +48,42 @@ export const LLMChatArea: React.FC<LLMChatAreaProps> = ({ config }) => {
...
+          chatHistory: chatMessages.current,
+          selectedTools: config.tools.map(tool => tool.name),
+          chatParams: config.extraLlmParams ? JSON.parse(config.extraLlmParams) : {}
+        })
+      })
+      if (response.ok) {
+        const responseText = await response.text()
+        chatMessages.current.push({ role: "user", content: userPrompt })
+        chatMessages.current.push({ role: "assistant", content: responseText })
+        displayMessage({ role: "assistant", content: responseText })
```
orpiske, line 81: The user's own messages are never displayed in the chat UI — only the assistant response is shown. The old code called `displayMessage({ role: "user", content: userPrompt })` before the request. This line should also display the user message:

```suggestion
        chatMessages.current.push({ role: "user", content: userPrompt })
        chatMessages.current.push({ role: "assistant", content: responseText })
        displayMessage({ role: "user", content: userPrompt })
        displayMessage({ role: "assistant", content: responseText })
```

apps/ui/admin/src/Pages/LLMChat/LLMChatArea.tsx
```diff
@@ -183,21 +96,9 @@ export const LLMChatArea: React.FC<LLMChatAreaProps> = ({ config }) => {
             size="lg"
             renderIcon={Send}
             iconDescription="Send"
-            disabled={isRunning}
             onClick={runPrompt}>
```
orpiske, line 99: The old code disabled the Send button via `disabled={isRunning}` while the request was in flight. With `isRunning` removed, users can fire multiple concurrent requests by clicking Send repeatedly. Consider adding back a loading state to disable the button during the request.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/chat/LlmChatResource.java
```diff
@@ -3,88 +3,185 @@
...
 
-            return response.body();
-        } catch (URISyntaxException ex) {
-            throw new WebApplicationException("Malformed base URL", ex, BAD_REQUEST);
-        } catch (IOException | InterruptedException ex) {
-            throw new WebApplicationException(INTERNAL_SERVER_ERROR);
+            ChatBot bot = AiServices.builder(ChatBot.class)
+                    .chatModel(model)
+                    .chatMemory(chatMemory)
+                    .systemMessage(systemPrompt)
```
orpiske, line 150: On subsequent messages, the frontend sends `systemPrompt: null`. This calls `.systemMessage(null)` on the AiServices builder. Depending on LangChain4j's handling, this may throw or silently drop the system prompt for subsequent turns. Worth verifying this behaves correctly — if LangChain4j does not accept null, this would need a conditional: only set `.systemMessage(...)` when `systemPrompt` is non-null.

orpiske (changes requested):
Thanks for this work, @rnetuka — the direction of moving LLM chat orchestration from the frontend to the backend via LangChain4j is a great step forward. It reduces API key exposure in browser network calls, centralizes tool execution, and opens up server-side chat memory. The `DefaultMcpBridge` bugfix (fresh `McpClient` per execution) and the `LLMSelect` UX improvement (names instead of URLs) are nice.

A few items need adjustment before merging. Inline comments below.

**General findings (not tied to specific lines):**

- **No tests for the new backend chat logic.** The PR adds significant new backend logic (multi-provider LLM integration, MCP tool provider, chat memory restoration, custom parameter mapping) but includes no tests. Per the project guidelines: "Major features should have tests." At minimum, `CustomLlmParameters.fromJson()` and `buildModel()` are easily unit-testable.

- **Dead code: `mcp.ts` is not deleted.** The PR removes the import of `McpClient` from `./mcp` in `LLMChatArea.tsx`, but `mcp.ts` itself is not deleted. It becomes dead code with no consumers.

**Note:** This is an OSS Helper conventions review. It does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/McpBridge.java
```diff
@@ -22,7 +22,7 @@ public interface McpBridge {
     List<RemoteToolReference> listTools(ForwardClient forwardClient) throws ServiceUnavailableException;
 
     Uni<ToolResponse> executeTool(
-            ForwardClient forwardClient, ToolManager.ToolArguments toolArguments, CallableReference toolReference);
+            String address, ToolManager.ToolArguments toolArguments, CallableReference toolReference);
```
orpiske, line 25: So, the bridge should be exclusive for MCP Forwards. It's not very clear to me why the change in this one.

orpiske: Maybe that's the one w/ the closed client issue?

rnetuka: @orpiske Yes, the ForwardClient is closed by the time the `executeTool`  is called. It needs to be created on-demand when the `executeTool` is being called.

orpiske: Got it. Then, it LGTM. I'll take another look at this one on Monday ... this week I've been busy w/ backports for Wanaku.

orpiske (approved):
LGTM, thanks

PR #1296, opened 2026-06-07
Refactor service tests to use mocks

### LINKED ISSUE #1193: Service catalog tests (Deploy/Expose/Init) run slow on Windows
## Description

The `ServiceDeployTest`, `ServiceExposeTest`, and `ServiceInitTest` tests in `apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/commands/service/` run slow on Windows.

They are currently disabled on Windows via `@DisabledOnOs(OS.WINDOWS)` as a workaround.

## Expected Behavior

These tests should run within a reasonable time on all platforms, including Windows.

## Additional Context

- All three tests use `@TempDir` and reflection-based field injection
- `ServiceInitTest` manipulates `user.dir` system property, which may interact poorly with Windows file locking
- The tests were disabled in the PR that added JUnit HTML test reports to CI

matheusandre1:
Closes: #1193 

Benchmark

Measured across 3 runs each, excluding JVM startup (test time only):

| Test Class | Tests | Before (avg) | After (avg) | Improvement |
|------------|-------:|-------------:|------------:|------------:|
| `ServiceDeployTest` | 3 | 0.327 s | 1.055 s | — |
| `ServiceExposeTest` | 9 | 0.301 s | 0.087 s | **3.5x** |
| `ServiceInitTest` | 4 | 0.143 s | 0.033 s | **4.3x** |

` ServiceDeployTest` The after-time is higher because the original tests were calling `cmd.call()` which swallows JLine terminal errors silently. The fixed tests call `doCall()` properly with mocked dependencies. The Mockito self-attach overhead (~0.7s one-time) accounts for most of the difference. On Windows, the original `call()` path was much slower (JLine native probing), so the net effect on Windows is a significant improvement.

My computer is average in terms of configuration, etc.

## Summary by Sourcery

Refactor service CLI command tests to invoke the interactive entrypoint with mocked dependencies and improve cross-platform reliability.

Tests:
- Update ServiceInit, ServiceExpose, and ServiceDeploy tests to use mocked Terminal and WanakuPrinter with doCall instead of call.
- Remove Windows-specific DisabledOnOs test exclusions by avoiding real JLine terminal usage and system directory changes.
- Normalize filesystem interactions in tests by using absolute paths and explicitly using UTF-8 for file I/O to make assertions stable across environments.

### DIFF
--- a/apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/commands/service/ServiceDeployTest.java
+++ b/apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/commands/service/ServiceDeployTest.java
@@ -3,30 +3,40 @@ package ai.wanaku.cli.main.commands.service;
 import java.io.File;
 import java.io.FileWriter;
 import java.lang.reflect.Field;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.Properties;
+import org.jline.terminal.Terminal;
+import ai.wanaku.cli.main.support.WanakuPrinter;
 
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.DisabledOnOs;
-import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.io.TempDir;
+import org.mockito.Mock;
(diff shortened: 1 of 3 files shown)


orpiske:
Nice! Thanks!

PR #1311, opened 2026-06-10
feat: Implement OpenTelemetry for distributed tracing and request ID propagation

### LINKED ISSUE #66: Traceability
We need to have better support for tracing incoming requests. 

As part of the connection, MCP clients and servers negotiate a session ID (i.e.; acquirable via `ToolManager.ToolArguments::requestId::asString` and `ResourceManager.ResourceArguments::requestId::asString`). I believe we can associate this with the requests all the way to the downstream services. This would allow us to match the requests to the invocations. 

Additionally, it would also be good to associate this ID with observability/tracing markers so we can trace this.

matheusandre1:
Closes: #66 
## Summary by Sourcery

Introduce distributed tracing and MCP request ID propagation across router, capabilities, and gRPC layers, and expose new LLM discovery endpoints.

New Features:
- Add OpenTelemetry-based tracing and Micrometer Prometheus metrics to the router backend and capabilities SDK.
- Introduce new chat endpoints and client APIs for listing allowed LLMs and their models.
- Provide an observability guide documenting tracing, logging, and request ID propagation.
- Add Jaeger and OpenTelemetry Collector services to local Docker Compose stacks for trace visualization.

Enhancements:
- Propagate W3C trace context and custom MCP request IDs through MCP HTTP clients, router bridges, and gRPC calls using interceptors and MDC.
- Extend gRPC proto contracts for tools, resources, and code execution to carry a request_id field for end-to-end correlation.
- Enrich logging formats across services to include traceId and requestId for easier log/trace correlation.
- Adjust proto SDK generation to stage files under target and copy missing protos into the source tree via a helper script.

Build:
- Update Maven POMs to include OpenTelemetry, Micrometer Prometheus, and logging dependencies and to run a script that copies SDK protos during generate-sources.

Deployment:
- Extend docker-compose configurations with OpenTelemetry Collector and Jaeger containers and configure services to export OTLP traces to the collector.

Documentation:
- Add detailed observability documentation covering tracing architecture, configuration, and request ID propagation.

## Summary by Sourcery

Introduce OpenTelemetry-based distributed tracing and MCP request ID propagation across router, capabilities, and gRPC layers, and integrate local tracing infrastructure.

New Features:
- Add OpenTelemetry tracing and Micrometer Prometheus metrics to the router backend and capabilities SDK.
- Introduce a request context abstraction to carry request and connection identifiers into spans and logs.
- Add Docker Compose services for an OpenTelemetry Collector and Jaeger with a shared collector configuration.
- Document the observability architecture, including tracing, logging, and request ID propagation.

Enhancements:
- Propagate MCP request IDs and W3C trace context through router bridges, gRPC interceptors, and MCP HTTP clients for end-to-end correlation.
- Extend tool, resource, and code execution gRPC requests to include a request_id field and wire it through router and capability delegates.
- Enrich router and capability logging formats to include traceId and requestId MDC values for improved log/trace correlation.
- Refine capability gRPC services and delegates to use a shared request context and tracing interceptor instead of directly implementing generated interfaces.

Build:
- Update Maven modules to include OpenTelemetry, Micrometer Prometheus, logging, and capabilities SDK dependencies, and align protobuf tooling versions in the parent POM.

Deployment:
- Configure router and capability services in Docker Compose to export OTLP traces to the local OpenTelemetry Collector and expose Jaeger for trace visualization.

Documentation:
- Add an observability guide describing tracing setup, span attributes, request ID propagation, and how to use the tracing stack.

### DIFF
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/CodeExecutionBridge.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/CodeExecutionBridge.java
@@ -94,16 +94,17 @@ public class CodeExecutionBridge implements CodeExecutorBridge {
      * @throws ServiceNotFoundException if no matching service is registered
      */
     @Override
-    public Iterator<CodeExecutionReply> executeCode(String engineType, String language, CodeExecutionRequest request) {
+    public Iterator<CodeExecutionReply> executeCode(
+            String engineType, String language, CodeExecutionRequest request, String requestId) {
         LOG.infof("Executing code (engine=%s, language=%s)", engineType, language);
 
-        // Resolve the service
         ServiceTarget service = resolveService(engineType, language);
 
-        // Build the gRPC request
-        ai.wanaku.core.exchange.v1.CodeExecutionRequest grpcRequest = buildGrpcRequest(engineType, language, request);
+        ai.wanaku.core.exchange.v1.CodeExecutionRequest grpcRequest =
+                buildGrpcRequest(engineType, language, request, requestId);
 
-        // Emit STARTED event
+        RequestIdContext.setContext(requestId, null);
+
         ToolCallEvent startedEvent = null;
         if (eventNotifier != null) {
             startedEvent = emitStartedEvent(engineType, language, service, request);
@@ -113,13 +114,10 @@ public class CodeExecutionBridge implements CodeExecutorBridge {
         final ToolCallEvent finalStartedEvent = startedEvent;
 
         try {
-            // Execute via transport
             Iterator<CodeExecutionReply> replyIterator = transport.executeCode(grpcRequest, service);
 
-            // Wrap the iterator to emit completion event when done or on close
             return new CloseableCodeExecutionIterator(replyIterator, finalStartedEvent, startTime, this);
         } catch (Exception e) {
-            // Emit FAILED event
             if (eventNotifier != null && startedEvent != null) {
                 long duration = Duration.between(startTime, Instant.now()).toMillis();
                 eventNotifier.emitFailedEvent(
@@ -129,6 +127,8 @@ public class CodeExecutionBridge implements CodeExecutorBridge {
                         duration);
             }
             throw e;
+        } finally {
+            RequestIdContext.clear();
         }
     }
 
@@ -253,7 +253,7 @@ public class CodeExecutionBridge implements CodeExecutorBridge {
      * Builds a gRPC CodeExecutionRequest from the SDK request.
      */
     private ai.wanaku.core.exchange.v1.CodeExecutionRequest buildGrpcRequest(
-            String engineType, String language, CodeExecutionRequest request) {
+            String engineType, String language, CodeExecutionRequest request, String requestId) {
 
         String uri = String.format("code-execution-engine://%s/%s", engineType, language);
 
@@ -263,7 +263,8 @@ public class CodeExecutionBridge implements CodeExecutorBridge {
                 ai.wanaku.core.exchange.v1.CodeExecutionRequest.newBuilder()
                         .setUri(uri)
                         .setCode(decodedCode)
-                        .setTimeout(request.getTimeout());
+                        .setTimeout(request.getTimeout())
+                        .setRequestId(requestId != null ? requestId : "");
 
         List<String> arguments = request.getArguments();
         if (arguments != null && !arguments.isEmpty()) {
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/InvokerBridge.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/InvokerBridge.java
@@ -1,5 +1,7 @@
 package ai.wanaku.backend.bridge;
 
+import jakarta.inject.Inject;
+
 import java.time.Duration;
 import java.time.Instant;
 import org.jboss.logging.Logger;
@@ -7,6 +9,7 @@ import io.quarkiverse.mcp.server.ToolManager;
 import io.quarkiverse.mcp.server.ToolResponse;
 import io.smallrye.mutiny.Uni;
 import io.smallrye.mutiny.infrastructure.Infrastructure;
+import io.vertx.core.Vertx;
 import ai.wanaku.backend.bridge.transports.grpc.GrpcTransport;
 import ai.wanaku.backend.common.ToolCallEvent;
 import ai.wanaku.backend.service.support.ServiceResolver;
@@ -32,6 +35,7 @@ public class InvokerBridge implements ToolsBridge {
     private final ServiceResolver serviceResolver;
     private final WanakuBridgeTransport transport;
     private final EventNotifier eventNotifier;
+    private final Vertx vertx;
 
     static class WanakuToolContext {
         ToolManager.ToolArguments arguments;
@@ -49,28 +53,16 @@ public class InvokerBridge implements ToolsBridge {
         }
     }
 
-    /**
-     * Creates a new InvokerBridge with the specified service resolver and transport.
-     *
-     * @param serviceResolver the resolver for locating tool services
-     * @param transport the gRPC transport for communication
-     */
-    public InvokerBridge(ServiceResolver serviceResolver, WanakuBridgeTransport transport) {
-        this(serviceResolver, transport, null);
-    }
-
-    /**
-     * Creates a new InvokerBridge with the specified service resolver, transport, and event notifier.
-     *
-     * @param serviceResolver the resolver for locating tool services
-     * @param transport the gRPC transport for communication
-     * @param eventNotifier the notifier for tool call events (nullable)
-     */
+    @Inject
     public InvokerBridge(
-            ServiceResolver serviceResolver, WanakuBridgeTransport transport, EventNotifier eventNotifier) {
+            ServiceResolver serviceResolver,
+            WanakuBridgeTransport transport,
+            EventNotifier eventNotifier,
+            Vertx vertx) {
         this.serviceResolver = serviceResolver;
         this.transport = transport;
         this.eventNotifier = eventNotifier;
+        this.vertx = vertx;
     }
 
     @Override
@@ -84,12 +76,23 @@ public class InvokerBridge implements ToolsBridge {
                             "Only local tool call references should be invoked by this executor"));
         }
 
+        String requestId = toolArguments.requestId().asString();
+        String connectionId = toolArguments.connection().id();
+
         return Uni.createFrom()
                 .item(() -> WanakuToolContext.create(toolArguments, ref))
                 .runSubscriptionOn(Infrastructure.getDefaultExecutor())
+                .invoke(ctx -> {
+                    if (vertx != null) {
+                        vertx.runOnContext(v -> RequestIdContext.setContext(requestId, connectionId));
+                    } else {
+                        RequestIdContext.setContext(requestId, connectionId);
+                    }
+                })
+                .invoke(ctx -> RequestIdContext.setToolName(ref.getName()))
                 .invoke(this::resolveService)
                 .invoke(ctx -> {
-                    ctx.request = InvokerToolExecutor.buildToolInvokeRequest(ref, toolArguments);
+                    ctx.request = InvokerToolExecutor.buildToolInvokeRequest(ref, toolArguments, requestId);
                     ctx.startTime = Instant.now();
                     if (eventNotifier != null) {
                         ctx.startedEvent =
@@ -105,7 +108,15 @@ public class InvokerBridge implements ToolsBridge {
 
                             LOG.debugf(failure, "Handling failure: %s", failure.getMessage());
                             return ToolResponse.error(failure.getMessage());
-                        }));
+                        }))
+                .onItemOrFailure()
+                .invoke((item, failure) -> {
+                    if (vertx != null) {
+                        vertx.runOnContext(v -> RequestIdContext.clear());
+                    } else {
+                        RequestIdContext.clear();
+                    }
+                });
     }
 
     private void emitCompleted(WanakuToolContext ctx, ToolResponse response) {
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/transports/grpc/RequestIdClientInterceptor.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/transports/grpc/RequestIdClientInterceptor.java
@@ -1,0 +1,46 @@
+package ai.wanaku.backend.bridge.transports.grpc;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+import org.jboss.logging.MDC;
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+
+@ApplicationScoped
+public class RequestIdClientInterceptor implements ClientInterceptor {
+
+    static final Metadata.Key<String> REQUEST_ID_KEY =
+            Metadata.Key.of("x-wanaku-request-id", Metadata.ASCII_STRING_MARSHALLER);
+
+    @Override
+    public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+            MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+
+        ClientCall<ReqT, RespT> delegate = next.newCall(method, callOptions);
+
+        String requestId = (String) MDC.get("requestId");
+        if (requestId != null && !requestId.isEmpty()) {
+            return withRequestId(delegate, requestId);
+        }
+
+        return delegate;
+    }
+
+    static <ReqT, RespT> ClientCall<ReqT, RespT> withRequestId(ClientCall<ReqT, RespT> delegate, String requestId) {
+        if (requestId != null && !requestId.isEmpty()) {
+            return new ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(delegate) {
+                @Override
+                public void start(Listener<RespT> responseListener, Metadata headers) {
+                    headers.put(REQUEST_ID_KEY, requestId);
+                    super.start(responseListener, headers);
+                }
+            };
+        }
+        return delegate;
+    }
+}
--- a/apps/wanaku-router-backend/src/main/resources/application.properties
+++ b/apps/wanaku-router-backend/src/main/resources/application.properties
@@ -216,6 +216,19 @@ quarkus.http.auth.permission.web.policy=authenticated
 
 # Logging configuration
 
+# OpenTelemetry configuration - disabled by default
+# Enable with -Dquarkus.otel.sdk.disabled=false and -Dquarkus.otel.exporter.otlp.endpoint=<endpoint>
+quarkus.otel.sdk.disabled=true
+%docker-compose.quarkus.otel.sdk.disabled=false
+%docker-compose.quarkus.otel.exporter.otlp.endpoint=http://otel-collector:4317
+quarkus.otel.exporter.otlp.traces.protocol=grpc
+quarkus.otel.resource.attributes=service.name=wanaku-router,service.version=${quarkus.application.version}
+quarkus.otel.traces.sampler=parentbased_always_on
+quarkus.otel.propagators=tracecontext,baggage
+
+# Enable MDC injection for trace correlation in logs
+quarkus.log.console.format=%d{yyyy-MM-dd HH:mm:ss} %-5p [%c] (%t) [traceId=%X{traceId}, requestId=%X{requestId}] %s%e%n
+
 quarkus.http.access-log.enabled=true
 quarkus.log.level=INFO
 quarkus.log.category."ai.wanaku".level=INFO
@@ -226,7 +239,7 @@ quarkus.log.category."io.quarkiverse.mcp".level=INFO
 
 %dev.quarkus.log.handler.file.wanaku-log.enabled=true
 %dev.quarkus.log.handler.file.wanaku-log.path=target/logs/wanaku.log
-%dev.quarkus.log.handler.file.wanaku-log.format=%d{yyyy-MM-dd HH:mm:ss} %-5p [%c] (%t) %s%e%n
+%dev.quarkus.log.handler.file.wanaku-log.format=%d{yyyy-MM-dd HH:mm:ss} %-5p [%c] (%t) [traceId=%X{traceId}, requestId=%X{requestId}] %s%e%n
 %dev.quarkus.log.category."ai.wanaku".handlers=wanaku-log
 %dev.quarkus.log.category."ai.wanaku".level=DEBUG
 %dev.quarkus.log.category."io.quarkus".handlers=wanaku-log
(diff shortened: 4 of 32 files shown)


core/core-exchange/src/main/proto/codeexecution.proto
```diff
@@ -73,6 +73,8 @@ message CodeExecutionRequest {
    * These environment variables will be set in the execution context.
    */
   map<string, string> environment = 8;
+
+  string request_id = 9;
```
orpiske, line 81: This and other .proto changes needs to happen in the Capabilities SDK

core/core-exchange/pom.xml
```diff
@@ -51,17 +51,39 @@
...
+            <executable>bash</executable>
+            <arguments>
+              <argument>${project.basedir}/copy-sdk-protos.sh</argument>
+              <argument>${project.build.directory}/proto-sdk</argument>
+              <argument>${project.basedir}/src/main/proto</argument>
+            </arguments>
+          </configuration>
+        </execution>
+      </executions>
+    </plugin>
```
orpiske: These changes shouldn't be needed if you do the proto changes on the SDK.

capabilities-quarkus-sdk/wanaku-capabilities-base/src/main/java/ai/wanaku/core/capabilities/provider/AbstractResourceDelegate.java
```diff
@@ -98,6 +98,10 @@ protected abstract List<String> coerceResponse(Object response)
 
     @Override
     public ResourceReply acquire(ResourceRequest request) {
+        String requestId = request.getRequestId();
+        if (requestId != null && !requestId.isEmpty()) {
+            org.jboss.logging.MDC.put("requestId", requestId);
+        }
         try {
```
orpiske, line 107: I'd like us to do this in a way that is transparent for the capabilities. 


Ideally, we would inject a bean (if observability is enabled) and then handle this in that bean scope.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/CodeExecutionBridge.java
```diff
@@ -100,13 +100,15 @@ public CodeExecutionBridge(
     public Iterator<CodeExecutionReply> executeCode(String engineType, String language, CodeExecutionRequest request) {
         LOG.infof("Executing code (engine=%s, language=%s)", engineType, language);
 
-        // Resolve the service
         ServiceTarget service = resolveService(engineType, language);
 
-        // Build the gRPC request
-        ai.wanaku.core.exchange.v1.CodeExecutionRequest grpcRequest = buildGrpcRequest(engineType, language, request);
+        String traceId =
```
orpiske: The same note about handling in a separate bean applies this. It is particularly important for the CCE, because the CCE is far from mature ... and we'd need to get as much as possible out of it because it's going to change significantly in the future.

orpiske (changes requested):
Hey @matheusandre1, I like the general strategy behind this. I think this one is going to be complex, though ... I did a first pass review, focusing the major items to focus. 

Thanks for this! :pray: Let's keep on top of this one

core/core-mcp-client/pom.xml
```diff
@@ -21,6 +21,20 @@
...
+            <artifactId>opentelemetry-api</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-context</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+             <artifactId>jboss-logging</artifactId>
```
orpiske: Minor formatting: the `<artifactId>jboss-logging</artifactId>` tag has an extra leading space compared to its sibling `<artifactId>` tags. Also, the file is missing a trailing newline.

deploy/docker-compose/otel-collector-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+        endpoint: 0.0.0.0:4317
+      http:
+        endpoint: 0.0.0.0:4318
+
+exporters:
+  otlp/jaeger:
+    endpoint: jaeger:4317
+    tls:
+      insecure: true
+  logging:
```
orpiske: The `logging` exporter with `loglevel` is deprecated in current OTel Collector versions (this PR pins `0.127.0`). The recommended replacement is the `debug` exporter with `verbosity`:

```yaml
debug:
  verbosity: basic
```

capabilities-quarkus-sdk/wanaku-capabilities-base/src/main/java/ai/wanaku/core/capabilities/common/TracingServerInterceptor.java
```diff
@@ -0,0 +1,87 @@
...
+    static final String SPAN_ATTR_REQUEST_ID = "wanaku.mcp.request_id";
+
+    @Override
+    public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
+            ServerCall<ReqT, RespT> call, Metadata headers, ServerCallHandler<ReqT, RespT> next) {
+
+        String requestId = headers.get(Metadata.Key.of("x-wanaku-request-id", Metadata.ASCII_STRING_MARSHALLER));
+
+        if (requestId != null && !requestId.isEmpty()) {
+            withRequestId(requestId, () -> {
```
orpiske: Question: `withRequestId()` here sets MDC, runs the lambda (which sets the span attribute), then immediately clears MDC. The `ServerCallListener` below re-sets MDC properly for each callback anyway. Is the `withRequestId()` wrapper intentional here, or would setting the span attribute directly (without the MDC set/clear cycle) be clearer?

orpiske (commented):
This PR includes changes to the REST API surface (renaming `/api/v1/chat/allowlist` → `/api/v1/chat/llms`, adding `/api/v1/chat/{llm}/models`) that are unrelated to tracing/observability. These update the OpenAPI specs, the generated TypeScript client, and the backend endpoints. Consider splitting them into a separate PR so the tracing feature can be reviewed and merged independently.

core/core-exchange/pom.xml
```diff
@@ -51,17 +51,39 @@
                                     <artifactId>capabilities-exchange</artifactId>
                                     <version>${wanaku-capabilities-sdk.version}</version>
                                     <classifier>sources</classifier>
-                                    <outputDirectory>${project.basedir}/src/main/proto</outputDirectory>
+                                    <outputDirectory>${project.build.directory}/proto-sdk</outputDirectory>
```
orpiske: I'm not sure why this change ... Can you elaborate about this one, please?

core/core-exchange/pom.xml
```diff
@@ -51,17 +51,39 @@
...
+              <argument>${project.build.directory}/proto-sdk</argument>
+              <argument>${project.basedir}/src/main/proto</argument>
+            </arguments>
+          </configuration>
+        </execution>
+      </executions>
+    </plugin>
+    <plugin>
+      <groupId>${quarkus.platform.group-id}</groupId>
+      <artifactId>quarkus-maven-plugin</artifactId>
```
orpiske: I think this is a left-over from the previous change ... so we shouldn't have it. I think core-exchange pom shouldn't have any change at all.

orpiske (changes requested):
@matheusandre1 this is shaping VERY NICELY! Really cool stuff here, man ... I'm eager to see it in 0.2.0 😱 🤩 

I also see that you did the proto changes on the Java SDK - which is the right way, so thanks for that too. I'll hold that one, until we sort out things here ... 

There's still a few things to adjust - as I pointed in a few more comments, but I like the overall direction 🙏  

Thanks!

orpiske:
FYI: I'll also do one round of automated review, to complement the manual one above.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/InvokerBridge.java
```diff
@@ -84,12 +84,17 @@ public Uni<ToolResponse> execute(ToolManager.ToolArguments toolArguments, Callab
                             "Only local tool call references should be invoked by this executor"));
         }
 
+        String requestId = toolArguments.requestId().asString();
+        String connectionId = toolArguments.connection().id();
+
         return Uni.createFrom()
                 .item(() -> WanakuToolContext.create(toolArguments, ref))
                 .runSubscriptionOn(Infrastructure.getDefaultExecutor())
+                .invoke(ctx -> RequestIdContext.setContext(requestId, connectionId))
```
orpiske: `RequestIdContext` uses `org.jboss.logging.MDC` which is thread-local, but this `.invoke()` runs inside a Mutiny reactive pipeline after `runSubscriptionOn(Infrastructure.getDefaultExecutor())`. The reactive chain can switch threads between operators, so MDC set here may not be visible in later stages, and `RequestIdContext.clear()` in `.onItemOrFailure()` might clear a different thread's MDC.

Consider using Vert.x duplicated context or Mutiny context propagation instead of raw thread-local MDC in reactive flows.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/CodeExecutionBridge.java
orpiske: Using the OTel trace ID as the MCP request ID conflates two semantically different identifiers. The trace ID is an OTel correlation token; the MCP request ID is a protocol-level identifier. Logs and traces will show a `requestId` that isn't actually an MCP request ID, which will confuse debugging.

If there's no MCP request ID available here, it's better to propagate null/empty and let the downstream handle it, rather than substituting a different type of ID.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/transports/grpc/RequestIdClientInterceptor.java
```diff
@@ -0,0 +1,38 @@
...
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.quarkus.grpc.GlobalInterceptor;
+
+@GlobalInterceptor
```
orpiske: Two concerns:

1. **Scope**: `@GlobalInterceptor` applies this to *all* gRPC client calls from the router, not just calls to capability services. If the router gains other gRPC client connections in the future, they'll all get the `x-wanaku-request-id` header injected.

2. **Dual propagation**: The request ID is now propagated both here (via gRPC metadata header) *and* inside the proto message body (`request_id` field). On the server side, `TracingServerInterceptor` reads the header while `AbstractToolDelegate`/`AbstractResourceDelegate` reads the proto field. This creates two parallel paths for the same data — consider picking one mechanism to avoid divergence.

apps/wanaku-router-backend/src/main/resources/application.properties
```diff
@@ -205,6 +205,16 @@ quarkus.http.auth.permission.web.policy=authenticated
 
 # Logging configuration
 
+# OpenTelemetry configuration
+quarkus.otel.exporter.otlp.endpoint=http://localhost:4317
```
orpiske: This hardcodes the OTel collector endpoint for all profiles. For Docker Compose it's overridden via env var, but for other deployments (Kubernetes/operator, local dev without a collector) there's no graceful fallback — the exporter will fail to connect and log errors.

Consider defaulting to `quarkus.otel.sdk.disabled=true` and enabling only in profiles/environments where a collector is available, or using a more defensive default.

orpiske (commented):
Additional review items not covered in prior rounds.

One general note: this PR adds 817 lines with zero test changes. `TracingServerInterceptor`, `RequestIdContext`, and `RequestIdClientInterceptor` are all unit-testable. The MDC thread-safety concern flagged on `InvokerBridge.java` especially warrants a test to verify correctness in reactive flows.

matheusandre1:
I made the corrections, but I need the pull request for the [SDK](https://github.com/wanaku-ai/wanaku-capabilities-java-sdk/pull/160)

orpiske:
> I made the corrections, but I need the pull request for the [SDK](https://github.com/wanaku-ai/wanaku-capabilities-java-sdk/pull/160)

Got it. I merged it, since it is harmless and should be backwards compatible.

orpiske:
I'm a bit overloaded w/ a PR of my own,  but I'll try to review this soon. It's on the top of my list.

orpiske:
Finally looking at this one ... I'll share updates soon/tomorrow

orpiske:
## CI Build Failure Report

I tested this PR by cherry-picking it onto a `ci-observability-pr-1311` branch (on my fork) to run the CI pipeline. The build failed on both `ubuntu-latest` and `ubuntu-24.04-arm` runners.

### Root Cause

The Javadoc plugin fails on `core-exchange` because the proto-generated package `ai.wanaku.core.exchange.v1` is not available during the `javadoc:jar` phase. The PR deletes the proto source files (`codeexecution.proto`, `healthprobe.proto`, `provision.proto`, `resourcerequest.proto`, `toolrequest.proto`) from `core/core-exchange/src/main/proto/` and replaces them with a protobuf-maven-plugin pipeline, but the generated classes are not present when Javadoc runs.

### Build Log (relevant excerpt)

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.12.0:jar
        (attach-javadocs) on project core-exchange: MavenReportException: Error while generating Javadoc:

[ERROR] .../core/core-exchange/src/main/java/ai/wanaku/core/exchange/ResourceAcquirerDelegate.java:13:
        error: package ai.wanaku.core.exchange.v1 does not exist
[ERROR] .../core/core-exchange/src/main/java/ai/wanaku/core/exchange/ProvisionAware.java:18:
        error: package ai.wanaku.core.exchange.v1 does not exist
[ERROR] .../core/core-exchange/src/main/java/ai/wanaku/core/exchange/InvocationDelegate.java:13:
        error: package ai.wanaku.core.exchange.v1 does not exist
[ERROR] .../core/core-exchange/src/main/java/ai/wanaku/core/exchange/HealthProbeDelegate.java:3:
        error: package ai.wanaku.core.exchange.v1 does not exist
[ERROR] 8 errors
```

### What was tested

- **Branch:** `ci-observability-pr-1311` (cherry-pick of `8233ac28` onto current `main`)
- **CI workflow:** `Build Main` (main-build.yml)
- **Run:** Both x86 and ARM failed with the same error

### Suggested Fix

The `core-exchange/pom.xml` may need to ensure the protobuf-maven-plugin runs before the `javadoc:jar` goal, or the Javadoc plugin should be configured to include the generated sources directory (e.g., `target/generated-sources/protobuf`) via `<additionalDependencies>` or `<sourcepath>`.

orpiske:
~Hm, that is not right. Checking what's going on ...~ 

Actually, it was. The test automation was correct.

core/core-exchange/pom.xml
```diff
@@ -60,8 +74,28 @@
...
+                            <goal>compile</goal>
+                            <goal>compile-custom</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+              <groupId>${quarkus.platform.group-id}</groupId>
+              <artifactId>quarkus-maven-plugin</artifactId>
             </plugin>
```
orpiske: These changes should not be here.

core/core-exchange/pom.xml
```diff
@@ -21,12 +21,26 @@
...
     </dependencies>
 
     <build>
+        <extensions>
+            <extension>
+                <groupId>kr.motd.maven</groupId>
+                <artifactId>os-maven-plugin</artifactId>
+                <version>${os-maven-plugin.version}</version>
+            </extension>
+        </extensions>
```
orpiske: No changes here either.

core/core-exchange/pom.xml
```diff
@@ -21,12 +21,26 @@
...
+
         <resources>
             <resource>
-                <directory>src/main/proto</directory>
+                <directory>${project.build.directory}/proto-sdk</directory>
```
orpiske: Invalid too.

core/core-exchange/pom.xml
```diff
@@ -21,12 +21,26 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-grpc</artifactId>
         </dependency>
+        <dependency> <!-- necessary for Java 9+ -->
+            <groupId>org.apache.tomcat</groupId>
+            <artifactId>annotations-api</artifactId>
+            <version>${annotations-api.version}</version>
+            <scope>provided</scope>
```
orpiske: Invalid too.

orpiske (changes requested):
Thanks ... but it needs a few more adjustments .... 

So, here's the problem: the build on core-exchange should not be modified. The _only_ changes on `core-exchange` are the refreshed `.proto` extracted automatically from the from the `capabilities-exchange`. In that module, nothing else should be modified.

orpiske:
## OpenShift Test Plan Results — PR #1311 (Attempt 2)

Tested commit `8c597ad57` (cherry-picked onto main via branch `ci-observability-pr-1311`).

### CI Build: PASS

Build succeeded on both `ubuntu-latest` (x86) and `ubuntu-24.04-arm` (ARM). Multi-arch manifests created and pushed successfully.

### Test Results

| Phase | Test | Result | Notes |
|-------|------|--------|-------|
| 4.1 | Prometheus metrics endpoint | **PASS** | `/q/metrics` returns 200 |
| 4.2 | JVM and HTTP metrics present | **PASS** | `jvm_` and `http_server_` metrics found |
| 7.1 | Logs contain `traceId=` | **PASS** | Structured log format working |
| 7.2 | Non-empty traceId in logs | **PASS** | e.g. `traceId=82cdf23c7a3737d49fc40a26d3adb72a` |
| 10.1 | Jaeger service discovery | **PASS** | `wanaku-router` found in Jaeger API |
| 10.3 | Traces found in Jaeger | **PASS** | 5 traces found for `wanaku-router` |
| 10.4 | Service names in traces | **PASS** | `wanaku-router` spans present |
| 10.5 | Custom `wanaku.mcp.*` attributes | **WARN** | Not tested — see blocker below |
| 10.6 | Parent-child span nesting | **WARN** | Not tested — see blocker below |
| 10.7 | Trace duration | **PASS** | 3.4ms — reasonable |

### Blocker: Noauth mode broken on this PR's image

The CI-built image cannot run in `noauth` mode on OpenShift. The error at startup:

```
Either 'jwks-path' or 'introspection-path' properties must be set when the discovery is disabled.
```

OIDC is a **build-time** Quarkus property. The `AuthConfigSource` sets `discovery-enabled=false` at runtime but the OIDC extension is already baked into the image. This prevents noauth deployments unless the image is re-augmented at startup (which the operator does not support).

When deploying **with auth** (Keycloak), the router starts correctly but the OIDC proxy (`AUTH_PROXY`) returns 404 on the token endpoint, so MCP tool invocations could not be authenticated end-to-end. This prevented testing:
- Phases 5-6 (tool invocation and resource acquisition via `wanaku mcp` CLI)
- Phase 8 (request ID correlation across router ↔ capability)
- Phase 10.5 (custom `wanaku.mcp.request_id` / `wanaku.mcp.tool_name` span attributes)
- Phase 10.6 (multi-span nesting across services)

### Additional Finding: Invalid Jaeger image tag

The `docker-compose-noauth.yml` and `docker-compose.yml` files specify `jaegertracing/jaeger:2.6` which does not exist on Docker Hub. The earliest v2 tag is `2.0.0` and the latest is `2.19.0`. During testing, `jaegertracing/jaeger:latest` was used instead.

### What was verified

The core OTEL instrumentation is working:
- Router generates W3C trace IDs and propagates them through logs
- Traces are successfully exported to the OTEL Collector and visible in Jaeger
- Prometheus metrics endpoint is functional with JVM and HTTP metrics
- Structured log format with `[traceId=..., requestId=...]` is confirmed

### What remains to be tested

Once the noauth/auth blocker is resolved:
- MCP tool invocation traces with `wanaku.mcp.request_id` and `wanaku.mcp.tool_name` attributes
- Cross-service span nesting (router → gRPC → capability)
- `requestId` correlation across router and capability logs
- Phase 9 (OTEL disabled by default negative test)

orpiske:
## OpenShift Test Plan Results — PR #1311 (Attempt 3)

Tested commit `8c597ad57` cherry-picked onto main, deployed on OpenShift with Keycloak auth, OTEL Collector, and Jaeger.

### CI Build: PASS

Both x86 and ARM builds succeeded. Multi-arch manifests pushed.

### Test Results

| Phase | Test | Result |
|-------|------|--------|
| 4.1 | Prometheus metrics endpoint | **PASS** |
| 4.2 | JVM/HTTP metrics present | **PASS** |
| 7.1 | `traceId=` and `requestId=` in router logs | **PASS** |
| 7.2 | Non-empty traceId in logs | **PASS** |
| 8.1 | requestId from MCP tool invocation in router logs | **PASS** (`requestId=3`, `requestId=5`) |
| 10.1 | `wanaku-router` in Jaeger service discovery | **PASS** |
| 10.3 | Traces found in Jaeger | **PASS** |
| 10.4 | Cross-service spans (router + capability) | **PASS** (trace `e3eec83429ae37bc712d10d38e112152`: 4 spans across `http` and `wanaku-router`) |
| 5.5 | `wanaku.mcp.*` span attributes | **FAIL** — see finding 1 |
| 8.2 | gRPC trace spans for tool invocations | **FAIL** — see finding 2 |

### Finding 1: MCP server tracing not enabled

The [Quarkus MCP Server observability docs](https://docs.quarkiverse.io/quarkus-mcp-server/dev/reference-observability.html) state that MCP tracing is off by default and requires:

```properties
quarkus.mcp.server.tracing.enabled=true
```

This property is not set in the PR. Without it, MCP operations (`tools/call`, `tools/list`, `initialize`, etc.) don't produce spans, and the attributes documented in the PR's `observability.md` (`wanaku.mcp.request_id`, `wanaku.mcp.tool_name`, `mcp.session.id`) are never emitted.

### Finding 2: gRPC calls are invisible to OpenTelemetry

The router's gRPC channels are created via raw `ManagedChannelBuilder.forTarget()` in `GrpcChannelManager`, bypassing Quarkus's managed gRPC client infrastructure. Quarkus auto-instruments only its own `@GrpcClient`-injected stubs, so these raw channels have no OTEL interceptor attached.

As a result, gRPC calls from the router to capability services (tool invocation, resource acquisition, provisioning) produce **no spans**. The `RequestIdClientInterceptor` added by this PR is a CDI `@ApplicationScoped` bean but is never registered on the `ManagedChannelBuilder` either.

To fix, either:
- Register an `OpenTelemetryClientInterceptor` (and the `RequestIdClientInterceptor`) on the channel via `builder.intercept(...)` in `GrpcChannelManager`
- Or migrate to Quarkus-managed `@GrpcClient` stubs, which get OTEL instrumentation automatically

### Finding 3: Jaeger image tag `2.6` does not exist

`docker-compose-noauth.yml` and `docker-compose.yml` specify `jaegertracing/jaeger:2.6`. This tag does not exist on Docker Hub — the earliest v2 tag is `2.0.0`, latest is `2.19.0`. During testing, `jaegertracing/jaeger:latest` was used.

### Finding 4: `service.name` resource attribute

The router appears in Jaeger under both `http` (from Vert.x auto-instrumentation) and `wanaku-router` (from the configured `quarkus.otel.resource.attributes`). The dual naming may cause confusion when searching for traces — some router spans land under `http` and others under `wanaku-router` depending on which instrumentation layer produced them.

orpiske (approved):
After the fix on the `core-exchange` this brought us a lot forward on implementation. Let's merge this and send the subsequent improvements on a separate PR.

orpiske:
@matheusandre1 great work on this improvement! Thanks!

Issue #773, opened 2026-02-11
[Persistence] Add Infinispan Remote persistence module


orpiske:
## Summary

Add a new `core-persistence-infinispan-remote` module to enable connection to external Infinispan server clusters for distributed deployments.

## Motivation

Currently, Wanaku uses **embedded Infinispan** (`core-persistence-infinispan`) for persistence. This works well for single-node deployments but has limitations:
- No shared state across multiple instances
- Data is local to each node
- Not suitable for clustered/distributed deployments

An **Infinispan Remote** module would enable:
- Connection to external Infinispan server clusters
- Shared state across all application instances
- Better scalability for production deployments
- Centralized data management

## Proposed Module Structure

```
core-persistence/
├── core-persistence-api/                   # Unchanged
├── core-persistence-infinispan/            # Embedded (existing)
└── core-persistence-infinispan-remote/     # NEW - Remote client
    └── src/main/java/ai/wanaku/core/persistence/infinispan/remote/
        ├── AbstractRemoteInfinispanRepository.java
        ├── AbstractLabelAwareRemoteInfinispanRepository.java
        ├── InfinispanRemote*Repository.java (6 concrete impls)
        ├── InfinispanRemotePersistenceConfiguration.java
        └── providers/
            └── InfinispanRemoteConfigurationProvider.java
```

## Key Implementation Changes

| Aspect | Embedded | Remote |
|--------|----------|--------|
| Cache Manager | `EmbeddedCacheManager` | `RemoteCacheManager` |
| Cache Type | `Cache<K,V>` | `RemoteCache<K,V>` |
| Dependency | `quarkus-infinispan-embedded` | `quarkus-infinispan-client` |
| Configuration | In-process | Hot Rod connection settings |
| Query | Embedded Ickle | Remote Ickle (via Hot Rod) |

## Dependencies

```xml
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-infinispan-client</artifactId>
</dependency>
```

## Configuration Properties

```properties
# Connection
quarkus.infinispan-client.hosts=localhost:11222
quarkus.infinispan-client.username=admin
quarkus.infinispan-client.password=secret

# TLS (optional)
quarkus.infinispan-client.trust-store=/path/to/truststore.jks
quarkus.infinispan-client.trust-store-password=secret
```

## Implementation Tasks

- [ ] Create module `core-persistence-infinispan-remote` with pom.xml
[...]

PR #1312, opened 2026-06-11
Add Infinispan Remote persistence module

matheusandre1:
draft #773 
## Summary by Sourcery

Introduce a new persistence layer backed by Infinispan Remote and integrate it into the core modules.

New Features:
- Add a core persistence API module defining generic repository and label-aware repository interfaces for Wanaku entities.
- Implement Infinispan Remote-backed repositories for resources, tools, prompts, namespaces, forward references, and data stores, including CDI wiring for use in the application.
- Add label expression parsing support to filter label-aware entities via a simple query language.

Enhancements:
- Configure Infinispan client caches, Protostream schemas, and marshallers to support indexed storage and querying of all persisted entities.
- Extend backend build configuration to version the Maven resources plugin used for UI distribution copying.

Tests:
- Add Quarkus-based integration tests for all Infinispan Remote repositories covering CRUD operations, name-based lookups, namespace allocation, and label-based filtering.
- Add unit tests for the label expression parser to validate supported syntax, operator precedence, and error handling.


matheusandre1:
Can you give me a hand here?

The error I'm accepting is because it needs to be modified in wanaku-capabilities-java-sdk, right? @orpiske

matheusandre1:
in error : Should these `@Indexed` / `@Keyword` / `@Text` annotations be added to **`wanaku-capabilities-java-sdk` / `capabilities-api`**, or is there another recommended location to place them?

orpiske:
@matheusandre1 FYI, it's on my TODO list to take a closer look at this, but I'm currently busy with a preparation for an event tonight.

orpiske (commented):
Thanks for the solid work on this, Matheus — the persistence API layer and Protobuf schemas are well structured, and the label expression parser with its test coverage is a nice addition.

A few observations and one question below.

core/core-persistence-infinispan-remote/src/main/resources/proto/content.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
```
orpiske, line 1: Every `.proto` file is duplicated identically in both `src/main/resources/META-INF/` and `src/main/resources/proto/`. The schemas are registered at runtime from `proto/` (via `AbstractWanakuSerializationContextInitializer`), while Infinispan auto-discovery reads from `META-INF/`. Having both copies means that if one is updated but the other isn't, serialization could break silently. Could one of the two locations be removed (or generated from the other)?

core/core-persistence-infinispan-remote/src/main/java/ai/wanaku/core/persistence/infinispan/remote/AbstractRemoteInfinispanRepository.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Override
+    public List<A> listAll() {
+        final RemoteCache<Object, A> cache = cacheManager.getCache(entityName());
+        return cache.values().stream().toList();
+    }
+
+    @Override
+    public boolean deleteById(K id) {
```
orpiske, line 47: `persist()`, `update()`, and `remove()` acquire the `ReentrantLock`, but `deleteById()`, `removeAll()`, and `removeByFields()` do not. If the lock is protecting against concurrent cache mutations, the unguarded methods could still race with guarded ones. Worth noting that the existing embedded Infinispan module may already have this same pattern — so this might be a deliberate choice or a pre-existing inconsistency to address separately.

core/core-persistence-infinispan-remote/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+        <dependency>
+            <groupId>ai.wanaku</groupId>
+            <artifactId>core-util</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>ai.wanaku.sdk</groupId>
+            <artifactId>capabilities-api</artifactId>
+        </dependency>
+    <dependency>
```
orpiske, line 29: Nit: from this line onward, the `<dependency>` blocks switch from 8-space/12-space indentation (matching the rest of the file) to 4-space/6-space. Looks like a copy-paste from a differently formatted source.

core/core-persistence-infinispan-remote/src/test/java/ai/wanaku/core/persistence/infinispan/remote/RemoteInfinispanTestProfile.java
```diff
@@ -0,0 +1,21 @@
...
+public class RemoteInfinispanTestProfile implements QuarkusTestProfile {
+
+    @Override
+    public Map<String, String> getConfigOverrides() {
+        return Map.of(
+                "quarkus.devservices.enabled", "true",
+                "quarkus.infinispan-client.devservices.enabled", "true",
+                "quarkus.infinispan-client.devservices.image-name", "infinispan/server:16.0",
+                "quarkus.infinispan-client.cache.tool.configuration-resource", "config/indexed-cache.json",
+                "quarkus.infinispan-client.cache.prompt.configuration-resource", "config/indexed-cache.json",
```
orpiske: The main `application.properties` uses `config/prompt-cache.json` for the `prompt` cache, but this test profile overrides it to `config/indexed-cache.json`. Is this intentional, or should the test profile also use `prompt-cache.json` to match production behavior?

orpiske (commented):
Github threw away my review - gotta admire how unreliable it has been lately. So, let's try again. 

I think the biggest challenge here is that this needs to be interchangeable. 

1. Embedded should be used for local / optionally for production 
2. Remove should be used in production (the default choice on OpenShift)

From a tactical PoV, I think it might be easier to start with #1310, but not necessarily mandatory ... just thinking out loud how to reduce the Infinispan footprint.

core/core-persistence-api/src/main/java/ai/wanaku/core/persistence/api/DataStoreRepository.java
```diff
@@ -0,0 +1,9 @@
+package ai.wanaku.core.persistence.api;
```
orpiske, line 1: Also, these files here, should not be duplicated ... if necessary, they should be moved from the existing ones so we don't lose history.

orpiske (changes requested):

matheusandre1:
> Github threw away my review - gotta admire how unreliable it has been lately. So, let's try again. 
> 
> I think the biggest challenge here is that this needs to be interchangeable. 
> 
> 1. Embedded should be used for local / optionally for production 
> 2. Remove should be used in production (the default choice on OpenShift)
> 
> From a tactical PoV, I think it might be easier to start with wanaku-ai/wanaku-barn#58, but not necessarily mandatory ... just thinking out loud how to reduce the Infinispan footprint. 

Thanks for the review and patience, I'll do issue wanaku-ai/wanaku#130, then we'll look at that pull request.

orpiske:
@matheusandre1 I think it's fine for you to continue on this one. I still need to have a chat with some of the Apicurio folks regarding my ideas and their plans. So you can continue, even though I may remove a class or two in a few releases in the future. I was just thinking out loud 😅

orpiske:
And thanks for the contribution ... this is an important one! 🙏

matheusandre1:
That approach didn't turn out well and isn't mature enough yet; I made a few mistakes. I'll be back soon with a better solution for this.

PR #1348, opened 2026-06-15
Fix #1336: stop persisting the LLM API key in browser localStorage

### LINKED ISSUE #1336: Stop persisting the LLM provider API key in browser localStorage
### Summary

The admin UI's LLM chat can persist the user-supplied provider API key in browser `localStorage` in
cleartext (the "store in local storage" toggle), where it is readable by any XSS or browser
extension. This is compounded by verbose MCP request/response logging in the `dev`/`perf` profiles.

### Remediation

Prefer `sessionStorage` or in-memory storage for the key; clearly warn users that `localStorage`
persists it across sessions; and avoid logging chat request bodies. Consider server-side key
vaulting instead of a client-supplied key.

### Affected

- `apps/ui/admin/src/Pages/LLMChat/*`
- chat request/response logging configuration

oscerd:
Closes #1336

The LLM chat UI persisted the entire config — including the provider API key — to `localStorage` in
cleartext when "Store in Local Storage" was enabled, where any XSS or browser extension on the page
could read it.

The API key is now excluded from what is stored: every write goes through a new `persistConfig()`
that strips `apiKey`. The key is kept in React state for the session only and must be re-entered
after a reload. The toggle label and a helper text on the API Key field make this explicit.
Provider/model/params/tools are still persisted as before. Also cleans up a pre-existing unused
catch binding in the same file so it lints clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Stop persisting LLM API keys in browser local storage and clarify how LLM settings are stored.

Bug Fixes:
- Exclude the LLM provider API key from configuration data written to browser local storage, keeping it only in session memory.

Enhancements:
- Introduce a dedicated config persistence helper that strips sensitive fields before saving and reuse it across the LLM setup UI.
- Update the LLM settings toggle label and API key field helper text to clearly communicate that API keys are never stored persistently.

### DIFF
--- a/apps/ui/admin/src/Pages/LLMChat/LLMSetup.tsx
+++ b/apps/ui/admin/src/Pages/LLMChat/LLMSetup.tsx
@@ -9,6 +9,7 @@ import {
   isConfigStoredInLocalStorage,
   LLM_CONFIG,
   LlmConfig,
+  persistConfig,
   STORE_IN_LOCAL_STORAGE
 } from "./config.ts"
 import {LLMSelect} from "./LLMSelect"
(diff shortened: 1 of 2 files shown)


orpiske:
Thanks, LGTM (c/c: @rnetuka). Let's get that one merged.

PR #1357, opened 2026-06-15
CLI fails to report any error #1288

### LINKED ISSUE #1288: CLI fails to report any error
My keycloak instance is in an https endpoint. 
Running this command:
```
wanaku admin credentials show --keycloak-url https://keycloak-wanaku.apps-crc.testing --admin-username admin --admin-password admin --client-id wanaku-service --show-secret --plain --verbose
```

Returns nothing.
I know the cause is the keycloak https certificate is self-signed.
Debugging, WanakuExceptionHandler handles the exception, but the `terminal` fails to actually prints anyting in the console.
It should be investigated.

claudio4j:
## Summary by Sourcery

Bug Fixes:
- Fix plain-mode terminal configuration so errors are properly reported via standard system input/output.


orpiske:
@claudio4j good catch, man! Thanks for that

orpiske (approved):

PR #1410, opened 2026-06-23
Fix: #1385  Qute content type configuration for plain text templates

### LINKED ISSUE #1385: Qute warns about unrecognized .tpl content type for deployment templates
During startup, Quarkus Qute logs WARN messages for .tpl files under templates/deployment/ because it cannot detect their content type, defaulting to application/octet-stream. Fix: register .tpl as a known content type via
    quarkus.qute.content-types.tpl=text/plain or rename to a recognized extension.

matheusandre1:
Fixes: #1385

## Summary by Sourcery

Build:
- Add Qute content type mapping for .tpl templates to text/plain in application configuration.


orpiske:
Cool, thanks!

Issue #1265, opened 2026-06-04
wanaku start local fails when using native binaries


orpiske:
## Description

`wanaku start local` fails with `AccessDeniedException: /Users/runner` when using native binaries built on CI.

## Root Cause

`RuntimeConstants.java` uses a static initializer to resolve `System.getProperty("user.home")`:

```java
static {
    WANAKU_HOME_DIR = System.getProperty("user.home") + File.separator + ".wanaku";
    WANAKU_CACHE_DIR = WANAKU_HOME_DIR + File.separator + "cache";
    WANAKU_LOCAL_DIR = WANAKU_HOME_DIR + File.separator + "local";
}
```

When GraalVM compiles the native image on GitHub Actions, this static initializer runs at **build time**, baking `/Users/runner` into the binary. At runtime on a different machine, the binary still tries to use `/Users/runner/.wanaku`, causing the `AccessDeniedException`.

## Stack Trace

```
ai.wanaku.capabilities.sdk.api.exceptions.WanakuException: java.nio.file.AccessDeniedException: /Users/runner
    at ai.wanaku.cli.main.support.Downloader.downloadFile(Downloader.java:45)
    at ai.wanaku.cli.runner.local.LocalRunner.downloadService(LocalRunner.java:223)
    at ai.wanaku.cli.runner.local.LocalRunner.deploy(LocalRunner.java:199)
    at ai.wanaku.cli.runner.local.LocalRunner.start(LocalRunner.java:78)
    at ai.wanaku.cli.main.commands.start.StartLocal.startWanaku(StartLocal.java:53)
    ...
Caused by: java.nio.file.AccessDeniedException: /Users/runner
    at sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:438)
    at java.nio.file.Files.createDirectories(Files.java:793)
    at ai.wanaku.cli.main.support.Downloader.downloadFile(Downloader.java:38)
```

## Fix

The static initializer in `RuntimeConstants` must be deferred to runtime. Options include:
- Use lazy initialization (compute on first access)
- Mark the class for `--initialize-at-run-time` in the native image config
- Use a method call instead of static fields

PR #1412, opened 2026-06-23
feat: centralize Wanaku home directory resolution and update related paths

### LINKED ISSUE #1398: Introduce a centralized `wanaku.home` property to replace scattered `user.home` usage
## Description

Currently, the code uses `System.getProperty("user.home")` in multiple places across the router, CLI, and capabilities to determine where to store files. This makes it difficult to point these components to a different directory.

## Proposal

Introduce a single `wanaku.home` system property (or configuration option) that:

- Defaults to `$HOME/.wanaku`
- Can be easily overridden via system property, environment variable, or configuration
- Is shared across the router, CLI, and capabilities

This centralizes the home directory resolution into one place, making it simpler to customize the storage location without modifying multiple code paths.

## Acceptance Criteria

- All usages of `System.getProperty("user.home")` for Wanaku file storage are replaced with the new `wanaku.home` resolution
- The default behavior (`$HOME/.wanaku`) remains unchanged
- The setting works consistently across router, CLI, and capabilities

matheusandre1:
Closes: #1398

## Summary by Sourcery

Centralize resolution of the Wanaku home directory across CLI, router backend, and capabilities components, and update configuration, docs, and tests to use the new abstraction and placeholders.

Enhancements:
- Introduce a core WanakuHome utility for resolving the Wanaku home directory via system property, environment variable, or user.home fallback.
- Update runtime constants, credential storage, service configuration, and Infinispan configuration to use WanakuHome-based path resolution and placeholder expansion instead of hardcoded user.home paths.
- Standardize configuration properties and logging paths to use a wanaku.home config placeholder for router and capabilities modules.

Documentation:
- Align configuration and persistence documentation to reference wanaku.home-based defaults instead of user.home-based paths.

Tests:
- Add unit tests for WanakuHome placeholder expansion and resolution behavior, and adjust existing CLI and capabilities tests to validate the new wanaku.home-based path handling.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/AuthCredentialStore.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/AuthCredentialStore.java
@@ -12,6 +12,7 @@ import java.nio.file.attribute.PosixFilePermission;
 import java.nio.file.attribute.PosixFilePermissions;
 import java.util.Properties;
 import org.jboss.logging.Logger;
+import ai.wanaku.core.util.WanakuHome;
 
 /**
  * A credential store specifically designed for CLI authentication.
@@ -22,7 +23,7 @@ public class AuthCredentialStore {
 
     private static final Logger LOG = Logger.getLogger(AuthCredentialStore.class);
 
-    private static final String DEFAULT_CREDENTIALS_FILE = "~/.wanaku/credentials";
+    private static final String DEFAULT_CREDENTIALS_FILE = WanakuHome.get() + File.separator + "credentials";
     private static final String API_TOKEN_KEY = "api.token";
     private static final String REFRESH_TOKEN_KEY = "refresh.token";
     private static final String AUTH_MODE_KEY = "auth.mode";
@@ -297,7 +298,7 @@ public class AuthCredentialStore {
     }
 
     private static URI resolveCredentialsPath(String credentialsPath) {
-        String resolvedPath = credentialsPath;
+        String resolvedPath = WanakuHome.expandPlaceholders(credentialsPath);
         if (resolvedPath.startsWith("~/")) {
             resolvedPath = System.getProperty("user.home") + resolvedPath.substring(1);
         }
--- a/apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/support/RuntimeConstantsTest.java
+++ b/apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/support/RuntimeConstantsTest.java
@@ -1,6 +1,7 @@
 package ai.wanaku.cli.main.support;
 
 import java.io.File;
+import ai.wanaku.core.util.WanakuHome;
 
 import org.junit.jupiter.api.Test;
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -10,8 +11,18 @@ class RuntimeConstantsTest {
 
     @Test
     void wanakuHomeDirUsesCurrentUserHome() {
-        String expected = System.getProperty("user.home") + File.separator + ".wanaku";
+        String expected = WanakuHome.get();
         assertEquals(expected, RuntimeConstants.wanakuHomeDir());
+    }
+
+    @Test
+    void wanakuHomeDirDefaultsToUserHomeDotWanaku() {
+        String userHome = System.getProperty("user.home");
+        assertNotNull(userHome);
+        String expected = userHome + File.separator + ".wanaku";
+        // When no wanaku.home system property is set, it should default to user.home/.wanaku
+        // This test validates the default behavior - in test env wanaku.home may not be set
+        assertEquals(expected, WanakuHome.get());
     }
 
     @Test
@@ -30,6 +41,6 @@ class RuntimeConstantsTest {
     void directoriesReflectSystemProperty() {
         String userHome = System.getProperty("user.home");
         assertNotNull(userHome);
-        assertEquals(userHome + File.separator + ".wanaku", RuntimeConstants.wanakuHomeDir());
+        assertEquals(WanakuHome.get(), RuntimeConstants.wanakuHomeDir());
     }
 }
--- a/apps/wanaku-router-backend/README.md
+++ b/apps/wanaku-router-backend/README.md
@@ -67,7 +67,7 @@ auth.server=http://localhost:8543
 quarkus.oidc.client-id=wanaku-mcp-router
 
 # Persistence
-wanaku.persistence.infinispan.base-folder=${user.home}/.wanaku/router/
+wanaku.persistence.infinispan.base-folder=${wanaku.home}/router/
 
 # MCP
 quarkus.mcp.server.traffic-logging.enabled=true
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/core/persistence/infinispan/providers/InfinispanConfigurationProvider.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/core/persistence/infinispan/providers/InfinispanConfigurationProvider.java
@@ -11,9 +11,10 @@ import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.configuration.cache.StorageType;
 import ai.wanaku.capabilities.sdk.api.exceptions.WanakuException;
+import ai.wanaku.core.util.WanakuHome;
 
 public class InfinispanConfigurationProvider {
-    @ConfigProperty(name = "wanaku.persistence.infinispan.base-folder", defaultValue = "${user.home}/.wanaku/router/")
+    @ConfigProperty(name = "wanaku.persistence.infinispan.base-folder", defaultValue = "${wanaku.home}/router/")
     String baseFolder;
 
     @ConfigProperty(name = "wanaku.persistence.infinispan.max-entries", defaultValue = "10000")
@@ -53,6 +54,6 @@ public class InfinispanConfigurationProvider {
     }
 
     private String resolveBaseFolder() {
-        return baseFolder.replace("${user.home}", System.getProperty("user.home"));
+        return WanakuHome.expandPlaceholders(baseFolder);
     }
 }
--- a/core/core-util/src/main/java/ai/wanaku/core/util/WanakuHome.java
+++ b/core/core-util/src/main/java/ai/wanaku/core/util/WanakuHome.java
@@ -1,0 +1,71 @@
+package ai.wanaku.core.util;
+
+import java.io.File;
+
+/**
+ * Centralized resolution of the Wanaku home directory.
+ * <p>
+ * The wanaku home directory is resolved in the following order:
+ * <ol>
+ *   <li>System property {@code wanaku.home}</li>
+ *   <li>Environment variable {@code WANAKU_HOME}</li>
+ *   <li>Default: {@code ${user.home}/.wanaku}</li>
+ * </ol>
+ */
+public class WanakuHome {
+
+    private static final String WANAKU_HOME_SYS_PROP = "wanaku.home";
+    private static final String WANAKU_HOME_ENV_VAR = "WANAKU_HOME";
+
+    private WanakuHome() {}
+
+    /**
+     * Returns the default Wanaku home directory path.
+     * <p>
+     * This is a method (not a static field) to avoid GraalVM native image
+     * build-time evaluation which would capture the build machine's home directory.
+     *
+     * @return the default path to the Wanaku home directory
+     */
+    private static String defaultWanakuHome() {
+        return System.getProperty("user.home") + File.separator + ".wanaku";
+    }
+
+    /**
+     * Returns the resolved Wanaku home directory path.
+     *
+     * @return the path to the Wanaku home directory
+     */
+    public static String get() {
+        String home = System.getProperty(WANAKU_HOME_SYS_PROP);
+        if (home != null && !home.isBlank()) {
+            return home;
+        }
+
+        home = System.getenv(WANAKU_HOME_ENV_VAR);
+        if (home != null && !home.isBlank()) {
+            return home;
+        }
+
+        return defaultWanakuHome();
+    }
+
+    /**
+     * Expands placeholders in a path string.
+     * <p>
+     * Currently supported placeholders:
+     * <ul>
+     *   <li>{@code ${wanaku.home}} - expands to the Wanaku home directory</li>
+     *   <li>{@code ${user.home}} - expands to the user's home directory</li>
+     * </ul>
+     *
+     * @param path the path string potentially containing placeholders
+     * @return the path with placeholders expanded
+     */
+    public static String expandPlaceholders(String path) {
+        if (path == null) {
+            return null;
+        }
+        return path.replace("${wanaku.home}", get()).replace("${user.home}", System.getProperty("user.home"));
+    }
+}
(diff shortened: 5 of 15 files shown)


core/core-util/src/main/java/ai/wanaku/core/util/WanakuHome.java
```diff
@@ -0,0 +1,60 @@
...
+ *   <li>System property {@code wanaku.home}</li>
+ *   <li>Environment variable {@code WANAKU_HOME}</li>
+ *   <li>Default: {@code ${user.home}/.wanaku}</li>
+ * </ol>
+ */
+public class WanakuHome {
+
+    private static final String WANAKU_HOME_SYS_PROP = "wanaku.home";
+    private static final String WANAKU_HOME_ENV_VAR = "WANAKU_HOME";
+    private static final String DEFAULT_WANAKU_HOME = System.getProperty("user.home") + File.separator + ".wanaku";
```
orpiske: This will cause native image to fail (i.e.: #1265)

orpiske (changes requested):
Thanks! Overall LGTM, but there is one issue that will cause native images to fail.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/core/persistence/infinispan/providers/InfinispanConfigurationProvider.java
```diff
@@ -11,9 +11,10 @@
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.configuration.cache.StorageType;
 import ai.wanaku.capabilities.sdk.api.exceptions.WanakuException;
+import ai.wanaku.core.util.WanakuHome;
 
 public class InfinispanConfigurationProvider {
```
orpiske, line 16: `WanakuHome.get()` already returns `$HOME/.wanaku`, so this default resolves to `$HOME/.wanaku/.wanaku/router/` — double `.wanaku`.

```suggestion
    @ConfigProperty(name = "wanaku.persistence.infinispan.base-folder", defaultValue = "${wanaku.home}/router/")
```

core/core-util/src/main/java/ai/wanaku/core/util/WanakuHome.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param path the path string potentially containing placeholders
+     * @return the path with placeholders expanded
+     */
+    public static String expandPlaceholders(String path) {
+        if (path == null) {
+            return null;
+        }
+        return path.replace("${wanaku.home}", get()).replace("${user.home}", System.getProperty("user.home"));
+    }
+}
```
orpiske, line 71: The class looks good — clean API, correct precedence. Two gaps:

1. **No unit tests** — should cover default resolution, sys prop override, `expandPlaceholders()` with `${wanaku.home}`, `${user.home}`, both, neither, and null.
2. Consider whether `DEFAULT_WANAKU_HOME` being `static final` (class-load time) could interact poorly with test harnesses that set `wanaku.home` system property after class loading. The `get()` method itself is fine since it checks the system property on each call.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/AuthCredentialStore.java
```diff
@@ -297,10 +298,7 @@ private static void restrictPermissions(Path path, String posixPermissions) {
     }
 
     private static URI resolveCredentialsPath(String credentialsPath) {
-        String resolvedPath = credentialsPath;
-        if (resolvedPath.startsWith("~/")) {
-            resolvedPath = System.getProperty("user.home") + resolvedPath.substring(1);
-        }
+        String resolvedPath = WanakuHome.expandPlaceholders(credentialsPath);
```
orpiske, line 301: This removes tilde (`~/`) expansion. `expandPlaceholders()` handles `${wanaku.home}` and `${user.home}` but not tildes. The default `DEFAULT_CREDENTIALS_FILE` was changed to an absolute path so the default case works, but `AuthCredentialStore(String credentialsPath)` accepts arbitrary paths — any caller passing `~/...` would get a broken URI.

Suggestion: either keep the tilde expansion as a fallback, or leave this file unchanged since it was out of scope per the issue.

```suggestion
        String resolvedPath = WanakuHome.expandPlaceholders(credentialsPath);
        if (resolvedPath.startsWith("~/")) {
            resolvedPath = System.getProperty("user.home") + resolvedPath.substring(1);
        }
```

apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/support/RuntimeConstantsTest.java
```diff
@@ -10,7 +11,7 @@ class RuntimeConstantsTest {
 
     @Test
     void wanakuHomeDirUsesCurrentUserHome() {
-        String expected = System.getProperty("user.home") + File.separator + ".wanaku";
+        String expected = WanakuHome.get();
```
orpiske, line 14: This test no longer verifies the actual resolved value — it just checks identity with `WanakuHome.get()`. Consider keeping the concrete assertion to validate the default path:

```suggestion
        String expected = System.getProperty("user.home") + File.separator + ".wanaku";
```

docs/troubleshooting.md
```diff
@@ -338,7 +338,7 @@ quarkus.log.category."ai.wanaku".level=DEBUG
 
 **Why this happens:**
 
-The Infinispan data directory defaults to `${user.home}/.wanaku/router/`. In container images, this resolves to `/home/jboss/.wanaku/router/`. Running the container with a different UID or mounting a host directory with incompatible permissions causes Infinispan's `SoftIndexFileStore` to fail.
+The Infinispan data directory defaults to `${wanaku.home}/.wanaku/router/`. In container images, this resolves to `/home/jboss/.wanaku/router/`. Running the container with a different UID or mounting a host directory with incompatible permissions causes Infinispan's `SoftIndexFileStore` to fail.
```
orpiske: Same double `.wanaku` issue — since `${wanaku.home}` already resolves to `$HOME/.wanaku`, this should be `${wanaku.home}/router/`.

```suggestion
The Infinispan data directory defaults to `${wanaku.home}/router/`. In container images, this resolves to `/home/jboss/.wanaku/router/`. Running the container with a different UID or mounting a host directory with incompatible permissions causes Infinispan's `SoftIndexFileStore` to fail.
```

docs/configurations.md
```diff
@@ -195,7 +195,7 @@ These `wanaku.router.health-check.*` properties control the periodic health prob
 
 | Property                                    | Description                                                                   |
 |---------------------------------------------|-------------------------------------------------------------------------------|
-| `wanaku.persistence.infinispan.base-folder` | Where to store Infinispan files (defaults to `${user.home}/.wanaku/router/`). |
+| `wanaku.persistence.infinispan.base-folder` | Where to store Infinispan files (defaults to `${wanaku.home}/.wanaku/router/`). |
```
orpiske: Same double `.wanaku` — should be `${wanaku.home}/router/`.

```suggestion
| `wanaku.persistence.infinispan.base-folder` | Where to store Infinispan files (defaults to `${wanaku.home}/router/`). |
```

apps/wanaku-router-backend/README.md
```diff
@@ -67,7 +67,7 @@ auth.server=http://localhost:8543
 quarkus.oidc.client-id=wanaku-mcp-router
 
 # Persistence
-wanaku.persistence.infinispan.base-folder=${user.home}/.wanaku/router/
+wanaku.persistence.infinispan.base-folder=${wanaku.home}/.wanaku/router/
```
orpiske: Same double `.wanaku` — should be `${wanaku.home}/router/`.

```suggestion
wanaku.persistence.infinispan.base-folder=${wanaku.home}/router/
```

orpiske (changes requested):
Thanks for tackling #1398, @matheusandre1! The overall approach is solid — `WanakuHome` in `core-util` with `get()` + `expandPlaceholders()` is exactly the right shape. There are two critical bugs to fix before this can merge, plus a few smaller items.

### Critical

1. **Double `.wanaku` in Infinispan paths** — `InfinispanConfigurationProvider` default resolves to `$HOME/.wanaku/.wanaku/router/` because `WanakuHome.get()` already includes `.wanaku`. Same bug in README, `docs/configurations.md`, `docs/troubleshooting.md`, and `docs/internals-persistence.md`. Compare with `WanakuServiceConfig.serviceHome()` which correctly uses `${wanaku.home}/services/`.

2. **`wanaku.home` not defined as a config source default** — `application.properties` files use `${wanaku.home}` as a SmallRye Config expression, but `wanaku.home` isn't defined in any config source. SmallRye resolves `${user.home}` because it's a Java system property; `wanaku.home` has no such backing. This will cause startup failures (at least in the `local` profile). Fix: add `wanaku.home=${user.home}/.wanaku` to the router and capabilities-base `application.properties`.

### Moderate

3. **No `WanakuHomeTest`** — The central utility class has no direct unit tests. Should cover: default resolution, system property override, precedence, and `expandPlaceholders()` with various inputs.

4. **Tilde expansion removed from `AuthCredentialStore`** — `resolveCredentialsPath()` originally expanded `~/` paths. `expandPlaceholders()` doesn't handle tildes. The default case works (changed to absolute path), but `AuthCredentialStore(String)` accepts arbitrary paths — a `~/...` path will break. This change was out of scope per the issue.

5. **`RuntimeConstantsTest` weakened** — Tests now verify `WanakuHome.get()` identity rather than the actual resolved value.

---

core/core-util/src/main/java/ai/wanaku/core/util/WanakuHome.java
orpiske: So, we can't use this either. Essentially, [trying to resolve the user's home dir in a static block in Quarkus will cause it to fail](https://github.com/quarkusio/quarkus/discussions/22772#discussioncomment-1940948). We we build native images during the release build, this will get fixed - in build time - to the $HOME of the github runner. 

Instead of having a static variable, what you want to have is a method named "`defaultWanakuHome`" that returns `System.getProperty("user.home") + File.separator + ".wanaku"` so it doesn't get "optimized" by GraalVM.

orpiske (changes requested):
Thanks, it's almost there! 

This would still lead to release build issues due to how GraalVM optimize building static variables.

core/core-util/src/main/java/ai/wanaku/core/util/WanakuHome.java
matheusandre1: Okay, I didn't know that, sorry.

matheusandre1:
The lint pipeline failure didn't happen in the files I modified; I don't get it.

orpiske:
## QE Verification Results — wanaku-home-resolution

> :robot: **Note:** This verification was executed by a coding agent. Results should be reviewed by a human before acting on them.

### Summary

| Metric | Count |
|--------|-------|
| Total tests | 44 |
| Passed | 42 |
| Failed | 2 |
| Skipped | 0 |
| Blocked | 0 |

### Overall Verdict: PARTIAL PASS

All critical functionality works correctly. The 2 failures are test plan expectation mismatches (not application bugs).

### Detailed Results

| Phase | Test ID | Test Name | Result | Notes |
|-------|---------|-----------|--------|-------|
| 0 | 0.1 | Verify required tools | PASS | Java 25, Maven 3.9.16, jq 1.8.2, curl 8.7.1 |
| 0 | 0.2 | Verify Java version >= 21 | PASS | |
| 0 | 0.3 | Create temporary test directories | PASS | |
| 0 | 0.4 | Verify no conflict with default home | PASS | |
| 1 | 1.1 | Verify build artifacts exist | PASS | Version 0.2.0-SNAPSHOT |
| 2 | 2.1 | Start local with default home | PASS | |
| 2 | 2.2 | Wait for router health (default) | PASS | Healthy on attempt 3 |
| 2 | 2.3 | Verify default Infinispan data directory | PASS | `~/.wanaku/router` exists |
| 2 | 2.4 | Verify default log file | PASS | `~/.wanaku/local/logs/wanaku-router.log` exists |
| 2 | 2.5 | Verify CLI local directory (default) | PASS | |
| 2 | 2.6 | Verify CLI cache directory (default) | FAIL | `cache` dir not created with `--local-dist` (test plan issue) |
| 2 | 2.7 | Stop default instance | PASS | |
| 2 | 2.8 | Verify default process is gone | PASS | |
| 3 | 3.1 | Start local with WANAKU_HOME | PASS | |
| 3 | 3.2 | Wait for router health (env var) | PASS | |
| 3 | 3.3 | Verify Infinispan data under custom home | PASS | |
| 3 | 3.4 | Verify log file under custom home | PASS | |
| 3 | 3.5 | Verify CLI local dir under custom home | PASS | |
| 3 | 3.6 | Verify CLI cache dir under custom home | FAIL | Same root cause as 2.6 |
| 3 | 3.7 | CLI connects with custom home | PASS | |
| 3 | 3.8 | Router persists data at custom location | PASS | 2318 Infinispan files |
| 3 | 3.9 | Verify log file has content | PASS | 79KB |
| 3 | 3.10 | Stop custom-home instance | PASS | |
| 3 | 3.11 | Verify custom-home process is gone | PASS | |
| 4 | 4.1 | Locate extracted router JAR | PASS | |
| 4 | 4.2 | Start router with -Dwanaku.home | PASS | |
| 4 | 4.3 | Wait for router health (sys prop) | PASS | |
| 4 | 4.4 | Verify Infinispan data under sys prop path | PASS | |
| 4 | 4.5 | Verify log file under sys prop path | PASS | |
| 4 | 4.6 | Stop system property router | PASS | |
| 4 | 4.7 | Verify sys prop process is gone | PASS | |
| 5 | 5.1 | Start router with both overrides | PASS | |
| 5 | 5.2 | Wait for router health (precedence) | PASS | |
| 5 | 5.3 | System property path has data (wins) | PASS | Precedence correct |
| 5 | 5.4 | Env var path has no data (loses) | PASS | Precedence correct |
| 5 | 5.5 | Log file under sys prop, not env var | PASS | |
| 5 | 5.6 | Stop precedence test router | PASS | |
| 5 | 5.7 | Verify precedence process is gone | PASS | |
| 6 | 6.1 | Credentials path uses custom home | PASS | |
| 6 | 6.2 | No credentials leak to default home | PASS | |
| 7 | 7.1 | Record default home state | PASS | 3867 files |
| 7 | 7.2 | Remove custom home directories | PASS | |
| 7 | 7.3 | Default home unchanged after cleanup | PASS | 3867 files unchanged |
| 8 | 8.1 | Unwritable WANAKU_HOME (negative) | PASS | Router failed to start as expected |
| 8 | 8.2 | Empty WANAKU_HOME falls to default | PASS | Blank check works correctly |
| 9 | 9.1 | Kill remaining test processes | PASS | |
| 9 | 9.2 | Remove temporary directories | PASS | |
| 9 | 9.3 | Unset test environment variables | PASS | |

### Key Findings

- **WanakuHome resolution works correctly** — system property > env var > default precedence verified end-to-end
- **All components respect the centralized path** — Infinispan data, log files, and CLI directories all redirect correctly
- **Backward compatibility preserved** — with no overrides, everything goes to `~/.wanaku/` as before
- **Negative tests pass** — unwritable paths fail properly, empty strings fall through to default
- **Cleanup isolation confirmed** — `~/.wanaku/` is untouched after removing custom home directories

The 2 FAIL results (tests 2.6 and 3.6) are the same root cause: the test plan expects a `cache/` directory to exist after `start local --local-dist`, but `--local-dist` bypasses the download mechanism that creates the cache. This is a test plan issue, not an application bug.

orpiske (approved):
Verified end-to-end: WanakuHome resolution (sys prop > env var > default) works correctly across router, CLI, and capabilities. All 42/44 tests pass; the 2 failures are test plan issues (cache dir not created in --local-dist mode), not application bugs. Backward compatibility preserved.

PR #1429, opened 2026-06-23
Add OpenAI chat tool configuration and routing files

### LINKED ISSUE #1184: Create a service template for camel-openai
## Summary

Create a service template for integrating with OpenAI via `camel-openai`.

This template should follow the existing service template conventions established in #1055.

matheusandre1:
Closes: #1184

## Summary by Sourcery

Add a new service template for routing chat completion requests to OpenAI models via Camel and exposing it as an MCP tool.

New Features:
- Introduce an OpenAI chat completion Camel route with configurable model and generation parameters.
- Expose the OpenAI chat completion route as an MCP tool with typed input properties for prompts and tuning parameters.
- Add service and catalog property files to register and configure the new OpenAI chat tool within the service templates.

### DIFF
--- a/services/service-templates/src/main/services/openai-chat-tool/index.properties
+++ b/services/service-templates/src/main/services/openai-chat-tool/index.properties
@@ -1,0 +1,7 @@
+catalog.dependencies.openai-chat=openai-chat/openai-chat.dependencies.txt
+catalog.description=Chat with OpenAI models (GPT-4, GPT-3.5, etc.)
+catalog.name=openai-chat-tool
+catalog.properties.openai-chat=openai-chat/service.properties
+catalog.routes.openai-chat=openai-chat/openai-chat.camel.yaml
+catalog.rules.openai-chat=openai-chat/openai-chat.rules.yaml
+catalog.services=openai-chat
--- a/services/service-templates/src/main/services/openai-chat-tool/openai-chat/openai-chat.camel.yaml
+++ b/services/service-templates/src/main/services/openai-chat-tool/openai-chat/openai-chat.camel.yaml
@@ -1,0 +1,55 @@
+- route:
+    id: openai-chat-completion
+    description: Chat completion using OpenAI models
+    from:
+      uri: direct:wanaku
+      steps:
+        - setHeader:
+            name: CamelOpenAI.operation
+            constant: "{{openai.operation}}"
+        - setHeader:
+            name: CamelOpenAI.model
+            constant: "{{openai.model}}"
+        - setHeader:
+            name: CamelOpenAI.temperature
+            constant: "{{openai.temperature}}"
+        - setHeader:
+            name: CamelOpenAI.maxTokens
+            constant: "{{openai.max.tokens}}"
+        - setHeader:
+            name: CamelOpenAI.topP
+            constant: "{{openai.top.p}}"
+        - setHeader:
+            name: CamelOpenAI.frequencyPenalty
+            constant: "{{openai.frequency.penalty}}"
+        - setHeader:
+            name: CamelOpenAI.presencePenalty
+            constant: "{{openai.presence.penalty}}"
+        - choice:
+            when:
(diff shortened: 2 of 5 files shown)


orpiske (changes requested):
Thanks for submitting this, @matheusandre1! The template structure, index.properties, rules, and dependencies all look good. I ran through the test plan and found two issues that prevent the route from loading:

### 1. `otherwise` placement in the Camel YAML DSL

In `openai-chat/openai-chat.camel.yaml`, the `otherwise` block is nested inside the `when` list (as a list item), but the Camel YAML DSL parser expects it to be a sibling of `when` at the `choice` level. The current structure:

```yaml
- choice:
    when:
      - simple: "..."
        steps: [...]
      - otherwise:    # <-- wrong: list item inside 'when'
          steps: [...]
```

Should be:

```yaml
- choice:
    when:
      - simple: "..."
        steps: [...]
    otherwise:        # <-- correct: sibling of 'when'
      steps: [...]
```

Without this fix, the CIC rejects the route with `Unsupported field: otherwise` and Camel starts with 0 routes.

### 2. Incorrect `openai` endpoint URI

The `to` URI on line 47 uses the API key as the path parameter:

```yaml
- to:
    uri: "openai:{{openai.api.key}}"
```

The `camel-openai` component expects the **operation** as the path parameter (`chat-completion` or `embeddings`), not the API key. The API key should be passed as a query parameter. Something like:

```yaml
- to:
    uri: "openai:chat-completion"
    parameters:
      apiKey: "{{openai.api.key}}"
      organization: "{{openai.organization}}"
```

Please check the [camel-openai component documentation](https://camel.apache.org/components/4.x/openai-component.html) for the exact URI format.

Both issues are blockers — the route cannot load or execute until they are fixed.

orpiske (commented):
I re-ran the test plan against the latest commit (438430b). The `otherwise` fix and the `openai:chat-completion` operation URI are now correct — thanks for addressing those!

However, the route still fails to load due to an additional issue:

### `organization` is not a valid `camel-openai` endpoint parameter

The Camel Integration Capability rejects the route with:

```
Failed to resolve endpoint: openai://chat-completion?apiKey=xxxxxx&organization=
There are 1 parameters that couldn't be set on the endpoint.
Unknown parameters=[{organization=}]
```

The `camel-openai` component does not expose `organization` as an endpoint parameter. You may need to either remove it or pass it via a header instead — please check the [camel-openai component documentation](https://camel.apache.org/components/4.x/openai-component.html) for the supported parameters.

Once this is fixed, the route should load and the end-to-end invocation tests can pass. Everything else in the test plan (template structure, listing, instantiation, catalog registration) works correctly.

orpiske (commented):
Re-ran the full test plan against the latest commit (7c29b34).

**All PR-specific tests pass:**

| Phase | Tests | Result |
|-------|-------|--------|
| 1: File structure | 1.1–1.7 | All PASS (7/7) |
| 2: REST API listing | 2.1–2.4 | All PASS (4/4) |
| 4: Instantiation | 4.1–4.5 | All PASS (5/5) |
| 5: Catalog verification | 5.1 | PASS |
| 6: CIC launch | 6.1 route loads, 6.2 tool registered | PASS |

The `otherwise` fix, operation URI fix, and `organization` removal all look correct. The route loads successfully in the Camel Integration Capability (total:1 route, `openai-chat-completion`).

**Remaining blockers for test 6.3 (MCP CLI tool invocation) are pre-existing infrastructure issues, not PR 1429 bugs:**

1. The router's health check fires immediately on CIC registration before the Camel context is ready, marking the tool as unhealthy. The periodic health check does not seem to recover the tool status afterward, so the tool never becomes visible via MCP `tools/list`. This needs to be tracked separately.
2. The tool description shows `{{tool.description}}` (unresolved placeholder) — minor, and may also be pre-existing.

From a template perspective, this PR is ready to merge once the `otherwise` fix is committed (which is already done in 7c29b34).

orpiske (approved):
Thanks @matheusandre1

PR #1462, opened 2026-06-27
fix: enhance reflection configuration with additional field and constructor parameter annotations

### LINKED ISSUE #1461: wanaku mcp resource list fails with native image deserialization error for McpResource
## Description

`wanaku mcp resource list` fails in native image mode with a Jackson deserialization error for `dev.langchain4j.mcp.client.McpResource`.

## Steps to Reproduce

```bash
wanaku mcp resource list --uri "http://<router>/public/mcp" --plain
```

## Expected Behavior

Resources from the MCP server should be listed.

## Actual Behavior

```
Cannot construct instance of `dev.langchain4j.mcp.client.McpResource`: cannot deserialize from Object value
(no delegate- or property-based Creator): this appears to be a native image, in which case you may need to
configure reflection for the class that is to be deserialized
```

## Root Cause

`McpResource` IS in `apps/wanaku-cli/src/main/resources/reflection-config.json` (line 12) with all field/method/constructor flags, but Jackson cannot introspect the `@JsonCreator` constructor parameter annotations (`@JsonProperty`) in native image mode. The reflection config exposes the class structure but NOT the constructor parameter metadata that Jackson needs for deserialization.

`McpTool` / `ToolSpecification` (used by `mcp tool list`) works because it uses a different construction path that doesn't rely on Jackson `@JsonCreator` deserialization.

## Suggested Fix

Update the reflection config entry for `McpResource` to include constructor parameter metadata, or add `quarkus.native.additional-build-args` for Jackson parameter names module.

matheusandre1:
Fixes: #1461

## Summary by Sourcery

Enhancements:
- Extend reflection-config.json to include extra field and constructor parameter annotations required at runtime.

### DIFF
--- a/apps/wanaku-cli/src/main/resources/reflection-config.json
+++ b/apps/wanaku-cli/src/main/resources/reflection-config.json
@@ -12,19 +12,23 @@
     "name": "dev.langchain4j.mcp.client.McpResource",
     "allDeclaredConstructors": true,
     "allPublicConstructors": true,
+    "allDeclaredFields": true,
+    "allPublicFields": true,
     "allDeclaredMethods": true,
     "allPublicMethods": true,
-    "allDeclaredFields": true,
-    "allPublicFields": true
+    "allConstructorParameterAnnotations": true,
+    "allPublicConstructorParameterAnnotations": true
   },
   {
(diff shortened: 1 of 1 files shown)


orpiske (approved):
Thanks!

PR #1463, opened 2026-06-27
fix: enhance OIDC authentication to support direct Keycloak login with realm option

### LINKED ISSUE #1447: wanaku auth login constructs wrong OIDC discovery URL when pointed at Keycloak directly
## Description

`wanaku auth login --auth-server <keycloak-url>` constructs the OIDC discovery endpoint as `<url>/q/oidc/.well-known/openid-configuration` (Quarkus OIDC reverse-proxy path). When `--auth-server` points to Keycloak directly, the correct path is `<url>/realms/<realm>/.well-known/openid-configuration`.

## Steps to Reproduce

```bash
wanaku auth login \
  --auth-server "http://keycloak-host" \
  --client-id "wanaku-service" \
  --realm "wanaku" \
  --username "admin" \
  --password
```

## Expected Behavior

When `--realm` is provided, the discovery URL should use the Keycloak-native path: `<auth-server>/realms/<realm>/.well-known/openid-configuration`.

## Actual Behavior

```
Authentication failed: Unable to download OpenID Provider metadata from
  http://keycloak-host/q/oidc/.well-known/openid-configuration: Status code 404
```

The `--realm` parameter is ignored for URL construction. The `forDiscovery()` method in `TokenEndpoint.java` (line 41) hardcodes `/q/oidc/`.

## Workaround

Point `--auth-server` to the router (port-forwarded) instead of Keycloak. The router's `/q/oidc/` reverse-proxy endpoint works correctly.

matheusandre1:
Fixes: #1447

## Summary by Sourcery

Enable the CLI OIDC authentication flow to work either via the router’s OIDC proxy or directly against a Keycloak realm, and document the new login options and behaviors.

New Features:
- Allow specifying a realm to construct a Keycloak-native OIDC discovery URL for direct authentication against Keycloak.
- Support username/password-based OAuth2 login in the CLI alongside token-based authentication.

Enhancements:
- Update the AuthLogin command to use realm-aware discovery URL selection and clarify CLI option semantics.
- Improve user documentation and troubleshooting guides to describe router-based vs direct Keycloak login flows and example commands.
- Extend test plans and add unit tests to validate realm-aware discovery URL construction and fallback behavior.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/auth/AuthLogin.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/auth/AuthLogin.java
@@ -13,7 +13,6 @@ import picocli.CommandLine;
 public class AuthLogin extends BaseCommand {
 
     private static final String DEFAULT_AUTH_SERVER = "http://localhost:8080";
-    private static final String DEFAULT_REALM = "wanaku";
     private static final String DEFAULT_CLIENT_ID = "admin-cli";
     private static final String DEFAULT_AUTH_MODE = "token";
 
@@ -40,8 +39,7 @@ public class AuthLogin extends BaseCommand {
 
     @CommandLine.Option(
             names = {"--realm"},
-            description = "Authentication realm",
-            defaultValue = DEFAULT_REALM)
+            description = "Authentication realm (e.g. 'wanaku'). When omitted, uses the router OIDC proxy.")
     private String realm;
 
     @CommandLine.Option(
@@ -77,13 +75,14 @@ public class AuthLogin extends BaseCommand {
                 config.setClientId(DEFAULT_CLIENT_ID);
                 config.setUsername(username);
                 config.setPassword(password);
-                config.setTokenEndpoint(TokenEndpoint.forDiscovery(serverUrl));
+                config.setTokenEndpoint(TokenEndpoint.forDiscovery(serverUrl, realm));
                 ServiceAuthenticator serviceAuthenticator = new ServiceAuthenticator(config);
 
                 credentialStore.storeApiToken(serviceAuthenticator.currentValidAccessToken());
                 credentialStore.storeRefreshToken(serviceAuthenticator.currentValidRefreshToken());
                 credentialStore.storeTokenExpiry(serviceAuthenticator.getTokenExpiryEpochSeconds());
                 credentialStore.storeClientId(DEFAULT_CLIENT_ID);
+                credentialStore.storeRealm(realm);
 
                 credentialStore.storeAuthMode("token");
                 credentialStore.storeAuthServerUrl(serverUrl);
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/security/TokenEndpoint.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/security/TokenEndpoint.java
@@ -32,13 +32,19 @@ public final class TokenEndpoint {
     }
 
     /**
-     * Constructs a token endpoint URL by appending the discovery OpenID Connect path to a base URL.
+     * Constructs a discovery URL using the Keycloak-native realm path when a realm is provided,
+     * or falls back to the Quarkus OIDC proxy path otherwise.
      *
      * @param baseUrl The base URL of the authentication server.
-     * @return The complete token endpoint URL.
+     * @param realm   The authentication realm, or {@code null} if not applicable.
+     * @return The complete discovery URL.
      */
-    public static String forDiscovery(String baseUrl) {
-        return stripTrailingSlash(baseUrl) + "/q/oidc/";
+    public static String forDiscovery(String baseUrl, String realm) {
+        String url = stripTrailingSlash(baseUrl);
+        if (realm != null && !realm.isBlank()) {
+            return url + "/realms/" + realm.strip();
+        }
+        return url + "/q/oidc/";
     }
 
     private static String stripTrailingSlash(String url) {
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -829,9 +829,19 @@ wanaku auth login --api-token <your-api-token>
 
 **Options:**
 
-- `--api-token <token>`: API token for authentication (required)
+- `--api-token <token>`: API token for authentication
 - `--auth-server <url>`: Authentication server URL (optional)
+- `--username <username>`: Username for password-based login
+- `--password <password>`: Password for password-based login (interactive)
+- `--realm <realm>`: Keycloak realm for direct Keycloak authentication (optional; when omitted, uses the router OIDC proxy)
+- `--client-id <client-id>`: OAuth2 client ID (default: `admin-cli`)
 - `--mode <mode>`: Authentication mode - `token` or `oauth2` (default: `token`)
+
+**Discovery URL behavior:**
+
+- When `--realm` is provided, the CLI constructs a Keycloak-native OIDC discovery URL: `<auth-server>/realms/<realm>/.well-known/openid-configuration`
+- When `--realm` is omitted (or blank), the CLI falls back to the Wanaku router OIDC proxy path: `<auth-server>/q/oidc/.well-known/openid-configuration`
+- This allows `wanaku auth login` to work directly against a Keycloak instance (e.g. `--auth-server http://keycloak-host --realm wanaku`) as well as through the router's OIDC proxy.
 
 **Example:**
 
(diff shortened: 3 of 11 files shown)


apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/auth/AuthLogin.java
```diff
@@ -77,7 +77,7 @@ public Integer doCall(Terminal terminal, WanakuPrinter printer) {
                 config.setClientId(DEFAULT_CLIENT_ID);
                 config.setUsername(username);
                 config.setPassword(password);
-                config.setTokenEndpoint(TokenEndpoint.forDiscovery(serverUrl));
+                config.setTokenEndpoint(TokenEndpoint.forDiscovery(serverUrl, realm));
```
orpiske, line 78: **Critical — `/q/oidc/` fallback path is unreachable (breaking change)**

The `--realm` option (line 41-45) declares `defaultValue = DEFAULT_REALM` (`"wanaku"`). Picocli resolves this before `doCall()` runs, so `realm` is **always** non-null and non-blank.

This means the guard in `forDiscovery`:
```java
if (realm != null && !realm.isBlank()) {
    return url + "/realms/" + realm.strip();
}
return url + "/q/oidc/";  // dead code — never reached
```

Users who currently run `wanaku auth login --auth-server http://localhost:8080` (the primary documented flow via the router's OIDC proxy) will now silently hit `http://localhost:8080/realms/wanaku` instead of `http://localhost:8080/q/oidc/`.

**Suggested fix** — remove the `defaultValue` from the `--realm` option so that omitting it leaves `realm` as `null`:
```java
@CommandLine.Option(
        names = {"--realm"},
        description = "Authentication realm (e.g. 'wanaku'). When omitted, uses the router OIDC proxy.")
private String realm;
```

If `DEFAULT_REALM` becomes unused after this change, consider removing the constant as well.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/security/TokenEndpoint.java
```diff
@@ -32,13 +32,19 @@ public static String fromBaseUrl(String baseUrl) {
...
      * @param baseUrl The base URL of the authentication server.
-     * @return The complete token endpoint URL.
+     * @param realm   The authentication realm, or {@code null} if not applicable.
+     * @return The complete discovery URL.
      */
-    public static String forDiscovery(String baseUrl) {
-        return stripTrailingSlash(baseUrl) + "/q/oidc/";
+    public static String forDiscovery(String baseUrl, String realm) {
+        String url = stripTrailingSlash(baseUrl);
+        if (realm != null && !realm.isBlank()) {
```
orpiske, line 44: The branching logic here is clean and well-structured — nice work. The only issue is that the caller (`AuthLogin`) currently guarantees `realm` is always `"wanaku"` due to the Picocli `defaultValue`, making the `else` branch dead code. Once the `defaultValue` is removed from `AuthLogin.java`, this will work as intended.

**Suggestion:** consider adding a unit test for this method covering all three paths (`realm` is null, blank, and a valid value). The `AuthCommandsTest.java` currently only covers `AuthLogout` and `AuthStatus` — there's no test coverage for `TokenEndpoint` or `AuthLogin` today.

docs/usage.md
```diff
@@ -829,10 +829,20 @@ wanaku auth login --api-token <your-api-token>
...
+- `--username <username>`: Username for password-based login
+- `--password <password>`: Password for password-based login (interactive)
+- `--realm <realm>`: Keycloak realm for discovery when authenticating directly against Keycloak (default: `wanaku`)
+- `--client-id <client-id>`: OAuth2 client ID (default: `admin-cli`)
 - `--mode <mode>`: Authentication mode - `token` or `oauth2` (default: `token`)
 
+**Discovery URL behavior:**
+
+- When `--realm` is provided, the CLI constructs a Keycloak-native OIDC discovery URL: `<auth-server>/realms/<realm>/.well-known/openid-configuration`
+- When `--realm` is omitted (or blank), the CLI falls back to the Wanaku router OIDC proxy path: `<auth-server>/q/oidc/.well-known/openid-configuration`
```
orpiske, line 843: This documentation says *"When `--realm` is omitted (or blank)..."* — but with the current `defaultValue = "wanaku"` in the code, `--realm` is never actually omitted. This describes behavior that cannot occur.

Once the `defaultValue` fix is applied in `AuthLogin.java`, this documentation will be accurate.

orpiske (changes requested):
Thanks for tackling this, @matheusandre1 — supporting direct Keycloak login is a real user need and the core logic in `forDiscovery` is clean and well-structured. However, there's a critical issue with the `defaultValue` on `--realm` that breaks the existing router-based authentication flow.

Please see the inline comments for details.

**Note:** This review checks against the project's rules, conventions, and behavioral correctness visible from the diff. It does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud for static analysis.

orpiske (commented):
Nice work on this PR, @matheusandre1! The fix is well-targeted and the test coverage is solid. A couple of things to flag:

### TokenRefresher gap (follow-up recommended)

`TokenRefresher.resolveIssuer()` at `TokenRefresher.java:108` still hardcodes `/q/oidc/` for discovery. When a user logs in via `--realm wanaku --auth-server http://keycloak-host`, the initial login will work correctly with this PR. However, when the token later expires and `TokenRefresher` tries to refresh, it reconstructs the URL using only the stored `authServerUrl` and appends `/q/oidc/` — which doesn't exist on Keycloak. The refresh will 404.

The realm would need to be persisted in the credential store and passed to `TokenRefresher` so it can use the same realm-aware logic. This can be tracked as a follow-up issue rather than blocking this PR, since login itself is strictly better than before.

docs/usage.md
```diff
@@ -829,10 +829,20 @@ wanaku auth login --api-token <your-api-token>
...
 **Options:**
 
-- `--api-token <token>`: API token for authentication (required)
+- `--api-token <token>`: API token for authentication
 - `--auth-server <url>`: Authentication server URL (optional)
+- `--username <username>`: Username for password-based login
+- `--password <password>`: Password for password-based login (interactive)
+- `--realm <realm>`: Keycloak realm for discovery when authenticating directly against Keycloak (default: `wanaku`)
+- `--client-id <client-id>`: OAuth2 client ID (default: `admin-cli`)
 - `--mode <mode>`: Authentication mode - `token` or `oauth2` (default: `token`)
```
orpiske, line 838: The code removed `defaultValue = DEFAULT_REALM` from the `--realm` option, so when `--realm` is omitted it is `null` and the CLI falls back to `/q/oidc/` (router proxy). This line says `(default: \`wanaku\`)` which is misleading — a user reading this might expect that omitting `--realm` still uses the Keycloak realm path.

Suggestion:
```suggestion
- `--realm <realm>`: Keycloak realm for direct Keycloak authentication (optional; when omitted, uses the router OIDC proxy)
```

matheusandre1:
> Nice work on this PR, @matheusandre1! The fix is well-targeted and the test coverage is solid. A couple of things to flag:
> ### TokenRefresher gap (follow-up recommended)
> 
> `TokenRefresher.resolveIssuer()` at `TokenRefresher.java:108` still hardcodes `/q/oidc/` for discovery. When a user logs in via `--realm wanaku --auth-server http://keycloak-host`, the initial login will work correctly with this PR. However, when the token later expires and `TokenRefresher` tries to refresh, it reconstructs the URL using only the stored `authServerUrl` and appends `/q/oidc/` — which doesn't exist on Keycloak. The refresh will 404.
> 
> The realm would need to be persisted in the credential store and passed to `TokenRefresher` so it can use the same realm-aware logic. This can be tracked as a follow-up issue rather than blocking this PR, since login itself is strictly better than before.
> 
> This review does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud.
> 
> _This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

Sorry for the delay; I'll look into this.

orpiske:
@matheusandre1 hey, no worries at all. This is a great improvement ... 🙏

orpiske (approved):
Tested the direct Keycloak login with `--realm` and router OIDC proxy login flows against a live Keycloak instance on OpenShift and a locally running router.

**21/23 tests passed** across 5 phases:
- Phase 12: Direct Keycloak login with `--realm` — all 8 tests PASS
- Phase 13: Router OIDC proxy login without `--realm` — all 5 tests PASS
- Phase 14: Realm persistence lifecycle (overwrite and clear) — both tests PASS
- Phase 15: Token refresh with stored realm — both tests PASS
- Phase 16: Negative tests — 4/6 PASS, 1 SKIP (pre-existing), 1 minor bug filed separately

Minor issue found: `--realm "   "` (whitespace-only) persists the blank value instead of treating it as null. Filed as a separate issue.

Issue #1264, opened 2026-06-03
Enable the CLI to handle TLS certificate 


claudio4j:
The CLI currently lacks options to handle the TLS on the server side:
* Skip TLS verification for development purposes.
* Supply a client certificate for server certificate validation.

If either the keycloak or wanaku server has TLS enabled endpoint, the CLI fails.

This doesn't cover the mutual TLS authentication.

orpiske:
Oh, good catch! Let's add this one to the backlog

claudio4j:
The work to add the `--insecure` to skip tls checks are in PR #1466.
To add support for the server certificates, it should be covered by usage of a trust store that contains the CA certificates of the target servers, this is mostly a configuration and documentation task than code. This is the next effort to complete this task.

claudio4j:
I will backport to 0.2.0 branch.

PR #1466, opened 2026-06-29
cli: Add --insecure cli parameter to trust all server certificates

claudio4j:
partial of #1264

## Summary by Sourcery

Add an --insecure CLI option to the Wanaku CLI to allow optional, development-only disabling of TLS certificate and hostname verification across HTTP, authentication, and admin flows, and centralize insecure HTTP/SSL handling.

New Features:
- Introduce a global --insecure CLI flag on base commands to trust all TLS certificates and disable hostname verification for development use.
- Add a convenience factory method on KeycloakAdminClient to create instances with optional insecure SSL configuration.
- Add HttpUtil and InsecureSSLHelper utilities to centralize creation of HttpClient and SSL contexts that support insecure TLS modes.
- Add OIDCHttpUtils utility to centralize OIDC metadata HTTP requests with optional insecure SSL support.

Enhancements:
- Wire the --insecure flag through authentication, token refresh, admin, and local runner flows so that all CLI HTTP interactions can consistently use insecure TLS when requested.
- Refine AuthLogin to respect the provided clientId and support insecure authentication flows.
- Adjust router backend build configuration to use an explicit maven-resources-plugin version.
- Improve terminal handling by always using the system terminal while keeping plain output mode support.
- Document how to skip certificate validation in the CLI and MCP server and clarify logging options in the FAQ.

Build:
- Pin the maven-resources-plugin version in the router backend Maven configuration.

Documentation:
- Extend the FAQ with instructions for disabling certificate validation for development and enabling verbose CLI logging.

Tests:
- Add tests for BaseCommand to verify secure vs insecure HttpClient configuration and extend AuthenticationInterceptor tests for the new constructor signature.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/BaseCommand.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/BaseCommand.java
@@ -1,11 +1,14 @@
 package ai.wanaku.cli.main.commands;
 
 import java.net.URI;
+import java.net.http.HttpClient;
 import java.util.concurrent.Callable;
 import org.jline.terminal.Terminal;
 import io.quarkus.rest.client.reactive.QuarkusRestClientBuilder;
 import ai.wanaku.cli.main.support.AuthenticationInterceptor;
+import ai.wanaku.cli.main.support.HttpUtil;
 import ai.wanaku.cli.main.support.WanakuPrinter;
+import ai.wanaku.cli.main.support.security.InsecureSSLHelper;
 import picocli.CommandLine;
 
 /**
@@ -46,9 +49,24 @@ public abstract class BaseCommand implements Callable<Integer> {
     protected boolean noAuth = false;
 
     @CommandLine.Option(
+            names = {"--insecure"},
+            description =
+                    "Skip certificate checks for the TLS connections (insecure, use only for development purposes)")
+    protected boolean insecure = false;
+
+    @CommandLine.Option(
             names = {"--plain"},
             description = "Route output through stdout so it can be captured by a parent process")
     boolean plain = false;
+
+    /**
+     * Creates an HttpClient instance, optionally configured for insecure SSL.
+     *
+     * @return a configured HttpClient
+     */
+    protected HttpClient createHttpClient() {
+        return HttpUtil.newHttpClient(insecure);
+    }
 
     /**
      * Initializes and configures the REST client for communicating with the Wanaku service.
@@ -64,7 +82,7 @@ public abstract class BaseCommand implements Callable<Integer> {
      * @throws IllegalArgumentException if the host URL is invalid or malformed
      * @throws NullPointerException if host is null
      */
-    protected static <T> T initService(Class<T> clazz, String host) {
+    protected <T> T initService(Class<T> clazz, String host) {
         return initService(clazz, host, null, false);
     }
 
@@ -72,7 +90,23 @@ public abstract class BaseCommand implements Callable<Integer> {
         return initService(clazz, host, this.authTokenOverride, this.noAuth);
     }
 
-    protected static <T> T initService(Class<T> clazz, String host, String tokenOverride, boolean noAuth) {
+    /**
+     * Initializes and configures the REST client for communicating with the Wanaku service.
+     *
+     * <p>Creates a properly configured Quarkus REST client builder with the specified host URL.
+     * The client is configured with appropriate timeouts and error handling for reliable
+     * communication with the Wanaku API endpoints.</p>
+     *
+     * @param <T> the type of the service interface
+     * @param clazz the Class object representing the service interface
+     * @param host the base URL of the Wanaku service API (must be a valid URI)
+     * @param tokenOverride to set the Bearer token in the http call
+     * @param noAuth to skip the keycloak authentication
+     * @return a configured Service instance ready for API calls
+     * @throws IllegalArgumentException if the host URL is invalid or malformed
+     * @throws NullPointerException if host is null
+     */
+    protected <T> T initService(Class<T> clazz, String host, String tokenOverride, boolean noAuth) {
         if (host == null) {
             throw new NullPointerException("Host URL cannot be null");
         }
@@ -81,10 +115,14 @@ public abstract class BaseCommand implements Callable<Integer> {
             QuarkusRestClientBuilder builder =
                     QuarkusRestClientBuilder.newBuilder().baseUri(URI.create(host));
 
+            if (insecure) {
+                builder.trustAll(true).hostnameVerifier(InsecureSSLHelper.getInsecureHostnameVerifier());
+            }
+
             if (!noAuth) {
                 AuthenticationInterceptor authInterceptor = tokenOverride != null
                         ? newAuthenticationInterceptor(tokenOverride)
-                        : new AuthenticationInterceptor();
+                        : new AuthenticationInterceptor(insecure);
                 builder.register(authInterceptor);
             }
 
@@ -94,7 +132,7 @@ public abstract class BaseCommand implements Callable<Integer> {
         }
     }
 
-    protected static <T> T initServiceIfNeeded(T existing, Class<T> clazz, String host) {
+    protected <T> T initServiceIfNeeded(T existing, Class<T> clazz, String host) {
         if (existing != null) {
             return existing;
         }
@@ -108,8 +146,8 @@ public abstract class BaseCommand implements Callable<Integer> {
         return initAuthenticatedService(clazz, host);
     }
 
-    private static AuthenticationInterceptor newAuthenticationInterceptor(String tokenOverride) {
-        return new AuthenticationInterceptor() {
+    private AuthenticationInterceptor newAuthenticationInterceptor(String tokenOverride) {
+        return new AuthenticationInterceptor(insecure) {
             @Override
             public void filter(jakarta.ws.rs.client.ClientRequestContext requestContext) {
                 requestContext
@@ -124,6 +162,10 @@ public abstract class BaseCommand implements Callable<Integer> {
         WanakuPrinter.setPlainMode(plain);
         try (Terminal terminal = WanakuPrinter.terminalInstance()) {
             WanakuPrinter printer = new WanakuPrinter(null, terminal);
+            if (insecure) {
+                printer.printWarningMessage(
+                        "WARNING: TLS certificate verification is disabled. This is insecure and should only be used for development.");
+            }
             return doCall(terminal, printer);
         } finally {
             WanakuPrinter.setPlainMode(false);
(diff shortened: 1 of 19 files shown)


apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/BaseCommand.java
```diff
@@ -65,14 +91,15 @@ public abstract class BaseCommand implements Callable<Integer> {
      * @throws NullPointerException if host is null
      */
     protected static <T> T initService(Class<T> clazz, String host) {
-        return initService(clazz, host, null, false);
+        return initService(clazz, host, null, false, false);
```
orpiske: **Critical:** This 2-parameter `initService` overload hardcodes `insecure = false`, which means the ~25+ commands that call `initService(SomeService.class, host)` (e.g., `ToolsList`, `ResourcesList`, `CapabilitiesList`, etc.) will silently ignore the `--insecure` flag.

A user running `wanaku tools list --insecure` will see the flag accepted but still get SSL errors — the most confusing possible behavior for a security flag.

Consider making this an instance method that reads `this.insecure`:

```suggestion
    protected <T> T initService(Class<T> clazz, String host) {
        return initService(clazz, host, null, false, this.insecure);
```

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/security/InsecureSSLHelper.java
```diff
@@ -0,0 +1,87 @@
...
+     * @return SSLParameters with endpoint identification disabled
+     */
+    public static SSLParameters createInsecureSSLParameters() {
+        SSLParameters sslParameters = new SSLParameters();
+        sslParameters.setEndpointIdentificationAlgorithm(null); // Disable hostname verification
+        return sslParameters;
+    }
+}
+
+// Made with Bob
```
orpiske: This looks like an AI tool attribution comment that should be removed before merging.

```suggestion
```

apps/wanaku-cli/src/main/java/ai/wanaku/cli/runner/local/LocalRunner.java
```diff
@@ -143,6 +148,19 @@ public LocalRunner(WanakuCliConfig config, LocalRunnerEnvironment environment) {
         this.routerReadinessUri = routerReadinessUri;
     }
 
+    private static HttpClient createHttpClient(boolean insecure) {
+        HttpClient.Builder builder = HttpClient.newBuilder();
+        if (insecure) {
+            try {
+                builder.sslContext(InsecureSSLHelper.createInsecureSSLContext())
```
orpiske: **Inconsistent error handling:** Here the failure is logged as a warning and the code silently falls back to a secure `HttpClient`. But the user explicitly asked for `--insecure` — falling back to secure means the subsequent SSL handshake will fail with a cryptic "certificate not trusted" error instead of the actual problem.

The same pattern appears in `AuthenticationInterceptor.isRouterAuthEnabled()`. In contrast, `BaseCommand.createHttpClient()` and `KeycloakAdminClient.create()` throw `RuntimeException`, which is the more helpful approach.

Consider failing consistently when `--insecure` was requested but cannot be configured.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/security/InsecureSSLHelper.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Utility for creating insecure SSL contexts that skip certificate verification.
+ * <p>
+ * WARNING: Only use for development/testing with self-signed certificates.
+ * This completely disables SSL certificate validation and hostname verification,
+ * making connections vulnerable to man-in-the-middle attacks.
+ * </p>
+ */
+public class InsecureSSLHelper {
```
orpiske, line 20: Minor suggestions for this utility class:

1. Add a private constructor to prevent accidental instantiation.
2. Consider narrowing `throws Exception` on `createInsecureSSLContext()` to `throws GeneralSecurityException` for better API specificity.

orpiske (changes requested):
Thank you for this feature, Claudio — the `--insecure` flag addresses a real developer need for working with self-signed certificates. The implementation shows solid understanding of the Java SSL stack: using `X509ExtendedTrustManager` and including `SSLParameters.setEndpointIdentificationAlgorithm(null)` are details many developers miss. The `BaseCommandTest` coverage is also well done.

However, there's one critical issue that would cause user confusion (see inline comment on `BaseCommand.java`), plus a few other items worth addressing.

### General Findings

**Code duplication:** The insecure `HttpClient` creation pattern (`if (insecure) { try { builder.sslContext(...) } catch ... }`) is repeated in 5+ locations (`BaseCommand`, `KeycloakAdminClient`, `LocalRunner`, `AuthenticationInterceptor`, plus Nimbus variants in `ServiceAuthenticator` and `TokenRefresher`). Consider adding a `createHttpClient(boolean)` helper to `InsecureSSLHelper` to consolidate.

**Scope:** The `WanakuPrinter` change (`system(!plainMode)` → `system(true)`) and `maven-resources-plugin` version pin appear unrelated to the `--insecure` feature. Consider splitting into separate commits for clarity.

**Note:** This review does not replace specialized tools like CodeRabbit, Sourcery, or SonarCloud.

orpiske:
> Scope: The WanakuPrinter change (system(!plainMode) → system(true)) and maven-resources-plugin version pin appear unrelated to the --insecure feature. Consider splitting into separate commits for clarity.

@claudio4j No need to worry about these ^. It's just part of my standard automated review.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/BaseCommand.java
claudio4j: fixed.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/runner/local/LocalRunner.java
claudio4j: fixed.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/security/ServiceAuthenticator.java
```diff
@@ -125,7 +147,7 @@ private static Issuer resolveIssuer(SecurityServiceConfig config) {
         }
     }
```
orpiske, line 122: This method calls `OIDCProviderMetadata.resolve(issuer)` (line 159) which internally creates its own HTTP connection using the Nimbus SDK's default SSL handling — the insecure SSL configuration applied to the manually-created `HTTPRequest` objects in `resolveIssuer()` and `requestToken()` doesn't carry over here.

When `--insecure` is set and Keycloak uses a self-signed cert, `resolveIssuer()` will succeed (insecure SSL is applied), but `OIDCProviderMetadata.resolve()` will fail with an SSL handshake error.

**Suggestion:** Replace the `OIDCProviderMetadata.resolve(issuer)` call with a manual `HTTPRequest` (with insecure SSL applied), parse the response, and extract the token endpoint URI — similar to how `resolveIssuer()` already works.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/security/TokenRefresher.java
```diff
@@ -130,7 +161,7 @@ private static Issuer resolveIssuer(String authServerUrl) {
         }
     }
```
orpiske, line 135: Same issue as in `ServiceAuthenticator` — `OIDCProviderMetadata.resolve(issuer)` (line 168) makes its own HTTP connection internally without insecure SSL. This will fail with self-signed certificates even when `--insecure` is set.

Same fix applies: manually fetch the OIDC metadata with an insecure-configured `HTTPRequest` and parse it, following the pattern already used in `resolveIssuer()` above.

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/admin/BaseAdminCommand.java
```diff
@@ -97,7 +97,7 @@ private String obtainAdminToken() {
                 .build();
 
         try {
-            HttpClient client = HttpClient.newHttpClient();
+            HttpClient client = insecure ? createHttpClient() : HttpClient.newHttpClient();
```
orpiske: Nit: this ternary is redundant — `createHttpClient()` already delegates to `HttpUtil.newHttpClient(insecure)`, which returns `HttpClient.newHttpClient()` when `insecure` is false. Simplifying to just `createHttpClient()` keeps things consistent with other call sites (e.g., `CapabilitiesWatch`).

```suggestion
            HttpClient client = createHttpClient();
```

apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/BaseCommand.java
```diff
@@ -45,11 +48,26 @@ public abstract class BaseCommand implements Callable<Integer> {
             description = "Disable authentication for this command")
     protected boolean noAuth = false;
 
+    @CommandLine.Option(
+            names = {"--insecure"},
+            description =
+                    "Skip certificate checks for the TLS connections (insecure, use only for development purposes)")
+    protected boolean insecure = false;
```
orpiske, line 55: **Suggestion:** Consider logging a visible warning when `--insecure` is active (e.g., `WARNING: TLS certificate verification is disabled. This is insecure and should only be used for development.`). This is common practice in tools like `curl -k` and `kubectl --insecure-skip-tls-verify`, and helps users notice when they've accidentally left the flag on.

orpiske (commented):
Great work on this PR — the `--insecure` flag is well-structured with clean centralization via `InsecureSSLHelper` + `HttpUtil`, proper `X509ExtendedTrustManager` usage, backward-compatible constructor overloads, and good test coverage. This is very close to merge-ready.

The one issue to address is the `OIDCProviderMetadata.resolve(issuer)` gap (see inline comments) — it will cause `--insecure` to fail for OIDC-authenticated setups with self-signed certificates, which is the primary use case for this flag.

orpiske:
@claudio4j hey, this is looking very good already! I did a manual review along with the automated one and I think the only critical issue to fix is the one with `OIDCProviderMetadata.resolve(issuer)`. Everything else is pretty much just a recommendation. 

Now, I will also run the test plan on top of this to see if we catch anything else and will post the results here.

orpiske:
## Smoke Test Results — CLI Start Local

Ran the `cli-start-local-smoke` test plan against this branch. **All tests passed** (failures are known pre-existing issues, not regressions from this PR).

| Metric | Count |
|--------|-------|
| Total tests | 20 |
| Passed | 17 |
| Known issues | 3 |

Tool registration, MCP tool invocation (end-to-end with httpbin), resource listing, prompt lifecycle (add/list/remove), admin UI reachability, and cleanup all verified successfully.

_This verification was executed by an AI agent and should be reviewed by a human._

orpiske:
@claudio4j thanks for the recent fixes. I think we are ready to merge. Please, can you rebase, so I can merge?

PR #1520, opened 2026-07-08
fix: throw exception when forward reference already exists

### LINKED ISSUE #1511: Forward add silently accepts duplicate names
## Description

Adding a forward with a name that already exists silently overwrites the existing forward without any warning or error.

## Steps to Reproduce

1. Start the local stack
2. Add a forward: `wanaku forwards add --name test-fwd --service http://localhost:8080/public/mcp/sse --namespace public`
3. Add another forward with the same name but different service: `wanaku forwards add --name test-fwd --service http://localhost:9090/public/mcp/sse --namespace public`

## Expected Behavior

Either:
- Reject the duplicate with an error (like `tools add` does with 409 Conflict), or
- Warn the user that an existing forward will be overwritten

## Actual Behavior

The second `add` silently overwrites the first forward. No warning or error is shown.

matheusandre1:
Fixes: #1511

### DIFF
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/forwards/ForwardsBean.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/forwards/ForwardsBean.java
@@ -144,8 +144,7 @@ public class ForwardsBean extends AbstractBean<ForwardReference> {
             // The input record is incomplete (i.e.: missing the ID, so we lookup the full record on the repository).
             final List<ForwardReference> references = forwardReferenceRepository.findByName(forwardReference.getName());
             if (!references.isEmpty()) {
-                LOG.warnf("Forward reference %s already exists", forwardReference.getName());
-                return;
+                throw EntityAlreadyExistsException.forName(forwardReference.getName());
             }
 
             registerForward(forwardReference);


matheusandre1:
> Hey - I've left some high level feedback:
> 
>     * Changing the behavior from a logged no-op to throwing EntityAlreadyExistsException alters the API contract; verify all upstream callers either expect this failure mode or are updated to handle the exception appropriately.
> 
>     * Consider whether the exception should include more context than just the name (e.g., the full ForwardReference or additional identifiers) to make debugging duplicate creation attempts easier.
> 
> 
> Prompt for AI Agents
> 
> ```
> Please address the comments from this code review:
> 
> ## Overall Comments
> - Changing the behavior from a logged no-op to throwing EntityAlreadyExistsException alters the API contract; verify all upstream callers either expect this failure mode or are updated to handle the exception appropriately.
> - Consider whether the exception should include more context than just the name (e.g., the full ForwardReference or additional identifiers) to make debugging duplicate creation attempts easier.
> ```
> 
> Sourcery is free for open source - if you like our reviews please consider sharing them ✨
> 
> 
> Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.

irrelant, There is no API contract breakage.

matheusandre1:
The callers already handle this exception.

PR #1521, opened 2026-07-08
refactor: error handling in DefaultMcpBridge

### LINKED ISSUE #1516: Forward path treats resources/list 'Method not found' as server unavailable
## Source

Reported in blog post: [Wanaku: Getting Started](https://www.the-main-thread.com/p/wanaku-getting-started)

## Description

When forwarding to the MCP memory server, Wanaku probes `resources/list`. The server replies with `Method not found` (it does not support resources), and Wanaku treats the entire upstream server as unavailable.

The author notes: "the current forward path is more demanding than 'any MCP server behind an SSE bridge'" — Wanaku effectively requires forwarded servers to support resource listing, even when the forward is only used for tools.

## Expected behavior

- A `Method not found` response to `resources/list` should be treated as "this server does not support resources," not "this server is down"
- Wanaku should gracefully degrade: if a server supports tools but not resources, tools should still work
- The capability probe should match what the forward is actually configured for (tools, resources, or both)

## Acceptance Criteria

- Handle `Method not found` responses from capability probes gracefully
- Do not mark a server as unavailable when only optional capabilities are missing
- Consider making the probe match the forward type (tool-only forwards should not fail because resources/list is unsupported)

matheusandre1:
Fixes: #1516

## Summary by Sourcery

Refine MCP bridge error handling for resource listing and forward registration to better tolerate unsupported methods while preserving consistent service unavailability semantics.

Bug Fixes:
- Treat MCP JSON-RPC 'method not found' errors on resources/list as an absence of resources instead of a hard failure.

Enhancements:
- Extract forward registration into dedicated helper methods for tools and resources to simplify control flow and cleanup.
- Standardize translation of MCP client failures during resource listing into ServiceUnavailableException, except when the method is not supported.

Tests:
- Add unit tests for DefaultMcpBridge.listResources covering unsupported methods, successful responses, and various error conditions.

### DIFF
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/forwards/ForwardsBean.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/forwards/ForwardsBean.java
@@ -29,6 +29,7 @@ import ai.wanaku.backend.core.mcp.util.LabelExpressionParser;
 import ai.wanaku.backend.core.persistence.api.ForwardReferenceRepository;
 import ai.wanaku.backend.core.persistence.api.WanakuRepository;
 import ai.wanaku.capabilities.sdk.api.exceptions.EntityAlreadyExistsException;
+import ai.wanaku.capabilities.sdk.api.exceptions.ServiceUnavailableException;
 import ai.wanaku.capabilities.sdk.api.exceptions.WanakuException;
 import ai.wanaku.capabilities.sdk.api.types.ForwardReference;
 import ai.wanaku.capabilities.sdk.api.types.LabelsAwareEntity;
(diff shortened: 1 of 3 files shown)


apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/api/v1/forwards/ForwardsBean.java
```diff
@@ -243,6 +238,29 @@ private void cleanupPartiallyRegistered(
         registeredRemoteToolsByForward.remove(nameNamespacePair);
     }
 
+    private void executeForwardRegistration(
+            ForwardReference forwardReference,
+            NameNamespacePair nameNamespacePair,
+            String address,
+            List<RemoteToolReference> locallyRegisteredTools,
+            Namespace ns) {
+        try (var forwardClient = ForwardClient.newClient(address)) {
```
orpiske, line 247: **Registration order reversed:** The original code registers resources first, then tools. This refactoring calls `registerTools()` before `registerResources()`, reversing the order. Since `cleanupPartiallyRegistered()` only cleans up tools (not resources), the original resources-first order is arguably safer. Consider swapping these two lines to preserve the original order.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/mcp/DefaultMcpBridge.java
```diff
@@ -115,6 +118,14 @@ public List<ResourceReference> listResources(ForwardClient forwardClient) throws
...
 
             return resourceRefs.stream().map(DefaultMcpBridge::remoteToLocal).collect(Collectors.toList());
+        } catch (McpException e) {
+            if (e.errorCode() == JSON_RPC_METHOD_NOT_FOUND) {
+                LOG.infof(
+                        "Remote MCP server at %s does not support resources/list, continuing without resources",
+                        forwardClient.address());
+                return List.of();
+            }
+            throw ServiceUnavailableException.forName(forwardClient.address());
```
orpiske: **Nit — original exception discarded:** This discards the `McpException` details (error code, message). Pre-existing pattern, but since this code is being touched, chaining the cause would improve debuggability:

```suggestion
            throw new ServiceUnavailableException(
                    String.format("Service is not available at %s", forwardClient.address()), e, false);
```

apps/wanaku-router-backend/src/test/java/ai/wanaku/backend/bridge/mcp/DefaultMcpBridgeTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    void listResources_throwsServiceUnavailableException_whenServerUnreachable() {
+        McpClient client = mock(McpClient.class);
+        when(client.listResources()).thenThrow(new McpException(-32600, "Internal error"));
+        ForwardClient forwardClient = new ForwardClient("http://localhost:9090", client);
+
+        assertThrows(
+                ai.wanaku.capabilities.sdk.api.exceptions.ServiceUnavailableException.class,
```
orpiske, line 53: **Nit:** Per JSON-RPC spec, error code `-32600` is "Invalid Request", not "Internal error" (which is `-32603`). The test behavior is correct regardless, but the message string is misleading.

```suggestion
        when(client.listResources()).thenThrow(new McpException(-32600, "Invalid request"));
```

orpiske (commented):
Thanks for this contribution, @matheusandre1! The core fix is well-targeted and the test coverage is solid. A few suggestions below.

**Consider applying the same `-32601` handling to `listTools()`:** Per the MCP spec, `tools/list` is also an optional capability. A resource-only server would hit the same problem in reverse. At minimum, a follow-up issue could track this; ideally the same pattern is applied symmetrically.

orpiske (approved):
LGTM, thanks!

PR #1535, opened 2026-07-11
feat: set default value for verified email in UsersAdd command

### LINKED ISSUE #1525: OIDC login fails with 'Account is not fully set up' when emailVerified=false
## Description

`wanaku auth login` fails with `ServiceAuthException: Account is not fully set up` when the Keycloak user has `emailVerified=false`.

## Steps to Reproduce

1. Create a user in the Keycloak `wanaku` realm using `wanaku admin users add`
2. Run `echo "password" | wanaku auth login --auth-server <ROUTER_URL> --username alice --password`
3. Observe error: `Account is not fully set up`

## Expected Result

Login should succeed, or provide a clear error message indicating which account setup step is missing.

## Actual Result

Login fails with:
```
ServiceAuthException: Account is not fully set up
  at ServiceAuthenticator.requestToken(ServiceAuthenticator.java:181)
```

## Root Cause

The `wanaku admin users add` command does not set `emailVerified=true` by default. The Keycloak realm configuration may require email verification as a default action.

## Workaround

Set `emailVerified=true` on the user via the Keycloak admin API before login.

## Suggested Fix

Either:
- Set `emailVerified=true` when creating users via `wanaku admin users add`, or
- Accept a `--email-verified` flag on the CLI command

## Environment

- Wanaku: latest (main branch)
- Keycloak: 26.6.1
- OpenShift: 4.18

---
*This issue was filed by a test agent during operator basic functionality verification.*

matheusandre1:
Closes: #1525

## Summary by Sourcery

Set the UsersAdd CLI command to treat email addresses as verified by default and document the requirement for verified emails for successful OIDC login.

New Features:
- Ensure the UsersAdd command defaults the email verified flag to true when creating users.

Documentation:
- Clarify in usage documentation that emails must be verified for OIDC login to succeed and describe the related Keycloak and CLI error behavior.

Tests:
- Add tests verifying that UsersAdd explicitly set to verified and with verified unset both result in a verified email by default.

### DIFF
--- a/apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/commands/users/UsersCommandsTest.java
+++ b/apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/commands/users/UsersCommandsTest.java
@@ -1,5 +1,6 @@
 package ai.wanaku.cli.main.commands.users;
 
+import java.lang.reflect.Field;
 import java.util.List;
 import java.util.Map;
 import org.jline.terminal.Terminal;
@@ -39,6 +40,27 @@ class UsersCommandsTest {
     }
 
     // ---- Happy-path tests ----
+
+    @Test
+    void usersAddDefaultsEmailVerifiedToTrue() throws Exception {
+        UsersAdd cmd = new UsersAdd(adminClient);
+        int result = cmd.doCall(terminal, printer);
+        assertEquals(EXIT_OK, result);
+        verify(adminClient).createUser(any(), any(), any(), any(), any(), any(), eq(true));
+        verify(printer).printSuccessMessage(any());
+    }
+
+    @Test
+    void usersAddHonorsNoVerifiedFlag() throws Exception {
+        UsersAdd cmd = new UsersAdd(adminClient);
+        Field verifiedField = UsersAdd.class.getDeclaredField("verified");
+        verifiedField.setAccessible(true);
+        verifiedField.setBoolean(cmd, false);
+        int result = cmd.doCall(terminal, printer);
+        assertEquals(EXIT_OK, result);
+        verify(adminClient).createUser(any(), any(), any(), any(), any(), any(), eq(false));
+        verify(printer).printSuccessMessage(any());
+    }
 
     @Test
     void usersAddHappyPath() throws Exception {
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/users/UsersAdd.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/users/UsersAdd.java
@@ -41,9 +41,8 @@ public class UsersAdd extends BaseAdminCommand {
     @CommandLine.Option(
             names = {"--verified"},
             description = "Mark the email address as verified (default: true)",
-            defaultValue = "true",
             negatable = true)
-    private boolean verified;
+    private boolean verified = true;
 
     public UsersAdd() {
         super();
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -1082,7 +1082,10 @@ wanaku admin users list
 wanaku admin users list --admin-username admin --admin-password admin
 
 # Create a new user (email, first-name, last-name are optional and default to username-based values)
-# By default, the email is marked as verified. Use --no-verified to leave it unverified.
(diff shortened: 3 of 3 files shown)


apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/commands/users/UsersCommandsTest.java
```diff
@@ -40,6 +41,27 @@ void setUp() {
...
     // ---- Happy-path tests ----
 
+    @Test
+    void usersAddDefaultsEmailVerifiedToTrue() throws Exception {
+        doNothing().when(adminClient).createUser(any(), any(), any(), any(), any(), any(), eq(true));
+        UsersAdd cmd = new UsersAdd(adminClient);
+        int result = cmd.doCall(terminal, printer);
+        assertEquals(EXIT_OK, result);
+        verify(printer).printSuccessMessage(any());
+    }
```
orpiske, line 51: The `doNothing().when(adminClient).createUser(...)` with `eq(true)` sets up a *stub*, but since `createUser` is a void method on a mock, it already does nothing by default. If `verified` were actually `false`, the stub wouldn't match — but Mockito would still do nothing (default behavior for void mocks), the method would succeed, and the test would pass.

To actually verify the default value, replace the stub with an explicit `verify()` assertion:

```suggestion
    @Test
    void usersAddDefaultsEmailVerifiedToTrue() throws Exception {
        UsersAdd cmd = new UsersAdd(adminClient);
        int result = cmd.doCall(terminal, printer);
        assertEquals(EXIT_OK, result);
        verify(adminClient).createUser(any(), any(), any(), any(), any(), any(), eq(true));
        verify(printer).printSuccessMessage(any());
    }
```

apps/wanaku-cli/src/test/java/ai/wanaku/cli/main/commands/users/UsersCommandsTest.java
```diff
@@ -40,6 +41,27 @@ void setUp() {
...
+    void usersAddHonorsNoVerifiedFlag() throws Exception {
+        doNothing().when(adminClient).createUser(any(), any(), any(), any(), any(), any(), eq(false));
+        UsersAdd cmd = new UsersAdd(adminClient);
+        Field verifiedField = UsersAdd.class.getDeclaredField("verified");
+        verifiedField.setAccessible(true);
+        verifiedField.setBoolean(cmd, false);
+        int result = cmd.doCall(terminal, printer);
+        assertEquals(EXIT_OK, result);
+        verify(printer).printSuccessMessage(any());
+    }
```
orpiske, line 63: Same issue as above — the `doNothing()` stub with `eq(false)` doesn't gate anything. The test passes regardless of what `verified` value is actually passed to `createUser`.

```suggestion
    @Test
    void usersAddHonorsNoVerifiedFlag() throws Exception {
        UsersAdd cmd = new UsersAdd(adminClient);
        Field verifiedField = UsersAdd.class.getDeclaredField("verified");
        verifiedField.setAccessible(true);
        verifiedField.setBoolean(cmd, false);
        int result = cmd.doCall(terminal, printer);
        assertEquals(EXIT_OK, result);
        verify(adminClient).createUser(any(), any(), any(), any(), any(), any(), eq(false));
        verify(printer).printSuccessMessage(any());
    }
```

orpiske (changes requested):
Thanks for the contribution, @matheusandre1! The intent here is solid — defaulting `emailVerified` to `true` is the right call for a smooth OIDC login experience, and the documentation update in `usage.md` is genuinely helpful.

The code change itself follows a valid Picocli pattern (field initialization for negatable booleans defaulting to `true`), which sidesteps the need to keep both `defaultValue` and `fallbackValue` in sync on the annotation.

However, there's one issue with the tests that needs to be addressed before merge — see the inline comments below.

**Additional suggestion:** The original code had `defaultValue = "true"` but was missing `fallbackValue = "true"`. Picocli docs state that negatable booleans defaulting to `true` require *both* annotation attributes, or can use field initialization instead (which is what this PR does). Mentioning this context in the PR description would help future maintainers understand the change.

Note: this review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

orpiske (approved):
Thanks!

PR #1540, opened 2026-07-13
chore: use Picocli ArgGroup for mutually exclusive and co-dependent CLI options

### LINKED ISSUE #1512: auth login without --username enters interactive mode instead of failing
## Description

When `wanaku auth login` is called with `--password` but without `--username`, the CLI enters interactive mode instead of reporting a missing required option.

## Steps to Reproduce

1. Run:
   ```
   echo "password" | wanaku auth login \
     --auth-server http://keycloak.example.com \
     --realm wanaku \
     --password \
     --plain
   ```
   (Note: `--username` is intentionally omitted)

## Expected Behavior

CLI should exit with non-zero status and print an error about missing `--username` when `--password` is provided.

## Actual Behavior

CLI exits with code 0, enters interactive mode, and prints "Interactive login mode" message.

## Additional Context

This is problematic for automation/scripting where missing a required flag should be a clear error, not a silent fallback to interactive mode.

orpiske:
## Summary

- Replace manual validation with declarative Picocli `@ArgGroup` annotations across the CLI
- `--token`/`--no-auth`: exclusive in `BaseCommand` (was silently undefined behavior)
- `--admin-username`/`--admin-password`: co-dependent in `BaseAdminCommand`
- `--name`/`--label-expression` and `--id`/`--label-expression`: exclusive via reusable `NameSelector` and `IdSelector` across 10 commands
- `--set`/`--get`/`--clear`: exclusive in `AuthToken` (was silently ignoring later options)
- `--api-token` vs `--username`+`--password`: exclusive with nested co-dependency in `AuthLogin`
- `--name`/`--clear-name`: exclusive in `NamespacesUpdate`
- Remove `LabelHelper.validateLabelExpression()` — Picocli now enforces these constraints at parse time with proper usage help

## Test plan

- [x] All existing unit tests pass (313 CLI tests, 0 failures)
- [ ] Verify `--token` and `--no-auth` cannot be used together
- [ ] Verify `--admin-username` requires `--admin-password` and vice versa
- [ ] Verify `--name`/`--label-expression` exclusivity in tools/resources/namespaces/datastores commands
- [ ] Verify `--set`/`--get`/`--clear` are mutually exclusive in `auth token`
- [ ] Verify `auth login` requires either `--api-token` or both `--username` and `--password`
- [ ] Verify `namespaces update --name` and `--clear-name` are mutually exclusive

## Summary by Sourcery

Enforce mutually exclusive and co-dependent CLI options across authentication, admin, and label-management commands using Picocli ArgGroups, while simplifying validation logic and centralizing selector handling.

New Features:
- Introduce reusable NameSelector and IdSelector argument groups to standardize name/id vs label-expression selection across tools, resources, namespaces, and datastores commands.
- Require explicit authentication mode in auth login, supporting either API token or username/password credentials as structured argument groups.
- Allow namespace name updates to be specified via a dedicated grouped option that enforces exclusivity between setting and clearing the name.

Bug Fixes:
- Prevent ambiguous auth token handling by making auth token override mutually exclusive with disabling authentication on all commands.
- Ensure auth token management subcommands (--set, --get, --clear) are mutually exclusive and no longer silently ignore additional options.
- Require both admin username and password when provided via CLI, avoiding partial admin credential configurations.
- Eliminate undefined combinations of auth login options by enforcing either API token or username/password, not both or neither.
- Prevent invalid combinations of id/name and label-expression selectors by delegating enforcement to Picocli ArgGroups instead of manual runtime checks.

Enhancements:
- Refactor label add/remove and resource/tool/datastore/namespaces removal commands to use shared selector types, reducing duplication and improving consistency.
- Simplify label expression validation by removing custom helper logic and relying on Picocli’s built-in argument group constraints.
- Update MCP commands to consume the new auth token override accessor, aligning them with the revised BaseCommand authentication options.

Tests:
- Update existing CLI tests to work with the new selector and name-update structures and remove now-obsolete validation tests tied to manual label expression checks.
- Adjust admin command tests to validate behavior with the new grouped admin credential options.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/BaseCommand.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/BaseCommand.java
@@ -38,15 +38,20 @@ public abstract class BaseCommand implements Callable<Integer> {
             description = "Display the help and sub-commands")
     private boolean helpRequested = false;
 
-    @CommandLine.Option(
-            names = {"--token"},
-            description = "Override authentication token for this command")
-    protected String authTokenOverride;
+    @CommandLine.ArgGroup(exclusive = true, multiplicity = "0..1")
+    protected AuthOptions authOptions;
 
-    @CommandLine.Option(
-            names = {"--no-auth"},
-            description = "Disable authentication for this command")
-    protected boolean noAuth = false;
+    static class AuthOptions {
+        @CommandLine.Option(
+                names = {"--token"},
+                description = "Override authentication token for this command")
+        String authTokenOverride;
+
+        @CommandLine.Option(
+                names = {"--no-auth"},
+                description = "Disable authentication for this command")
+        boolean noAuth = false;
+    }
 
     @CommandLine.Option(
             names = {"--insecure"},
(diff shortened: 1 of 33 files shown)


orpiske:
Good catch — added an explicit `if (operation.clearToken)` guard around the clear logic. While the `@ArgGroup(multiplicity = "1")` guarantees exactly one option is set at parse time, the explicit check makes the code self-documenting and resilient to future changes. Fixed in 6f345e0.

PR #1562, opened 2026-07-14
fix #1557: validate required tool parameters before forwarding to capability

### LINKED ISSUE #709: Operator Deployment Support
# Operator Deployment Support

## Objective
Enable the Wanaku operator to deploy and manage the Camel Code Execution Engine capability as a Kubernetes resource.

## Requirements
- Define Custom Resource Definition (CRD) for Camel Code Execution Engine
- Implement operator reconciliation logic for deploying the capability
- Support configuration of security controls (allowlists/blocklists) via CRD
- Support configuration of dependency caching strategy via CRD
- Handle lifecycle management (create, update, delete) of capability instances
- Provide status reporting in CRD status field
- Support configuring resource limits and requests
- Enable integration with existing Wanaku service discovery

## Affected Files/Modules
- Wanaku operator codebase (CRD definitions)
- Wanaku operator reconciliation controllers
- Camel Code Execution Engine deployment manifests
- Helm charts (if applicable)

## Notes
- CRD should expose security configuration (component/endpoint/route filters) as declarative spec
- Consider supporting both in-cluster and remote deployment modes
- Operator should validate CRD configurations before deployment
- Status field should report deployment state, active routes, and health checks
- Integration with Infinispan for template storage may require configuration injection

### LINKED ISSUE #1557: Router does not validate required tool parameters before forwarding to capability
## Description

When a tool is registered with `--required` parameters, the router does not validate that those parameters are present in MCP tool call requests before forwarding to the capability. The `required` metadata is stored and advertised in the MCP tool schema, but never enforced server-side.

## Steps to Reproduce

1. Register a tool with a required parameter:
   ```bash
   wanaku tools add --host http://localhost:8080 --name test-tool --namespace public \
     --description "Test" --uri "https://httpbin.org/get?q={query}" --type http \
     --property "query:string,Search query" --required query
   ```
2. Invoke the tool via MCP **without** the required parameter:
   ```bash
   wanaku mcp tool --uri http://localhost:8080/public/mcp/sse --name test-tool --plain
   ```
3. Observe: the call is forwarded to the capability instead of being rejected

## Expected

The router should return an MCP error response indicating the required parameter `query` is missing, before forwarding to the downstream capability.

## Current Behavior

The router forwards the request directly. The capability receives the call without the parameter and may produce unexpected results or errors.

## Context

- `InvokerBridge.execute()` builds and forwards the gRPC request without any input schema validation
- The `required` field is correctly stored in `InputSchema` and visible via `tools show --plain`
- MCP clients (e.g., Claude) use the schema to fill in required params, but raw MCP calls bypass this

## Found by

QE run 2026-07-14, test plan `archetype-full-lifecycle`, test 5.5.

orpiske:
Closes #1557

## Summary
- Added server-side validation of required tool parameters in the router before forwarding to downstream capabilities
- Missing required parameters now return an MCP error response instead of being silently forwarded

## Test plan
- [ ] Register a tool with `--required query`
- [ ] Call it via MCP without `query` parameter — should get error
- [ ] Call it via MCP with `query` parameter — should succeed

## Summary by Sourcery

Validate MCP tool invocations against required input parameters before forwarding to downstream capabilities.

Bug Fixes:
- Reject tool invocations that omit or provide null/blank values for schema-declared required parameters instead of forwarding them silently.

Tests:
- Add unit tests covering required-parameter validation for present, missing, blank, null, and multiple-required argument scenarios.

### DIFF
--- a/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/InvokerToolExecutor.java
+++ b/apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/InvokerToolExecutor.java
@@ -1,11 +1,14 @@
 package ai.wanaku.backend.bridge;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 import org.jboss.logging.Logger;
 import io.quarkiverse.mcp.server.ToolManager;
+import ai.wanaku.capabilities.sdk.api.types.InputSchema;
 import ai.wanaku.capabilities.sdk.api.types.Property;
 import ai.wanaku.capabilities.sdk.api.types.ToolReference;
 import ai.wanaku.core.exchange.v1.ToolInvokeRequest;
@@ -47,6 +50,9 @@ public final class InvokerToolExecutor {
      */
     static ToolInvokeRequest buildToolInvokeRequest(
             ToolReference toolReference, ToolManager.ToolArguments toolArguments, String requestId) {
+
+        // Validate required parameters before processing
+        validateRequiredParameters(toolReference, toolArguments.args());
 
         // Filter out metadata and auth args before converting to string map
         Map<String, Object> filteredArgs = filterOutReservedArgs(toolArguments.args());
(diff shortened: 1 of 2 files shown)


apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/InvokerToolExecutor.java
```diff
@@ -181,6 +187,40 @@ private static void requireNonNullValue(Map.Entry<String, Property> entry, Objec
...
+        List<String> missing = new ArrayList<>();
+        for (String param : required) {
+            Object value = args.get(param);
+            if (value == null || (value instanceof String s && s.isBlank())) {
+                missing.add(param);
+            }
+        }
+
+        if (!missing.isEmpty()) {
+            throw new IllegalArgumentException(
```
orpiske, line 221: How the backend will handle this type of exception? What will the CLI see?

orpiske (commented):
Added some notes.

apps/wanaku-router-backend/src/main/java/ai/wanaku/backend/bridge/InvokerToolExecutor.java
orpiske: The `IllegalArgumentException` thrown by `validateRequiredParameters` is caught by `InvokerBridge.execute()`'s Mutiny `.onFailure().recoverWithItem()` chain (InvokerBridge.java, lines 105-110), which converts it to `ToolResponse.error(failure.getMessage())`. The CLI receives the error message through the MCP protocol response — for example: `"Missing required parameter(s): query"`.

orpiske (commented):
Thanks for the review feedback.

**Null guard for `args`**: Addressed in b1f9fab. `validateRequiredParameters` now treats a null `args` map as empty, and a new test covers this case.

**Folding into existing validation flow**: The existing `requireNonNullValue` and `evalValue` methods serve a different purpose — they validate that properties defined in the input schema have a non-null value (either as a default or from the arguments), and they run during header/value extraction. `validateRequiredParameters` validates the MCP `required` field contract — checking that all parameters listed in `inputSchema.required` are present and non-blank *before* any processing begins. These are complementary checks at different stages, so keeping them separate is the cleaner approach.

PR #1570, opened 2026-07-15
fix #1510: tools import shows spurious Internal Server Error messages

### LINKED ISSUE #1510: tools import shows spurious Internal Server Error messages during import
## Description

When importing tools generated from an OpenAPI spec, the CLI displays "Internal Server Error" / "Generic error" messages during the import process. Despite these messages, the command exits with code 0 and the tools are successfully imported.

## Steps to Reproduce

1. Create an OpenAPI spec file
2. Generate tools: `wanaku tools generate spec.yaml`
3. Import the generated tools: `wanaku tools import --host http://localhost:8080 generated-tools.json`

## Expected Behavior

Clean import with success messages only.

## Actual Behavior

"Internal Server Error" messages appear in output, but the import succeeds (exit code 0, tools are registered).

## Additional Context

The error messages are confusing to users since the operation actually succeeds. The server may be returning 500 for some intermediate operations while the overall import completes.

orpiske:
## Summary

- Guard against `null` description in `ToolsHelper.registerTool()` and `ResourceHelper.expose()` — the MCP library's `setDescription()` calls `Objects.requireNonNull()`, which throws NPE when OpenAPI-generated tools have no description. Defaults to empty string.
- Track import failures in `ToolHelper.importToolset()` and return `EXIT_ERROR` from `ToolsImport` and `ToolsGenerate` when any tool fails to import.
- Handle HTTP 409 (already exists) gracefully during import — prints a skip message instead of treating it as an error.

## Test plan

- [x] `mvn verify` passes for `wanaku-router-backend` (69 tests)
- [x] `mvn verify` passes for `wanaku-cli`
- [ ] Manual test: import tools with `null` description — should succeed without 500
- [ ] Manual test: re-import existing tools — should print skip messages, exit 0
- [ ] Manual test: import with a mix of valid and server-error tools — should exit 1 with failure count

## Summary by Sourcery

Handle null tool and resource descriptions and improve CLI tool import error reporting and exit codes.

Bug Fixes:
- Avoid null pointer exceptions when registering tools or exposing resources that have no description.
- Treat HTTP 409 conflicts during tool import as non-fatal, printing a skip message instead of failing the import.
- Return non-zero exit codes from ToolsImport and ToolsGenerate when any tool import fails.

Enhancements:
- Track and report the number of failed tool imports in CLI commands to provide clearer feedback.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/tools/ToolsGenerate.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/tools/ToolsGenerate.java
@@ -99,7 +99,7 @@ public class ToolsGenerate extends BaseCommand {
                 System.err.print("\n\nImporting toolset...");
                 int failures = importToolset(toolReferences, host);
                 if (failures > 0) {
-                    System.err.printf("Import completed with %d failure(s)%n", failures);
+                    printer.printErrorMessage(String.format("Import completed with %d failure(s)", failures));
                     return EXIT_ERROR;
                 }
                 System.err.print("Done.");
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/tools/ToolsGenerate.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/tools/ToolsGenerate.java
@@ -97,7 +97,11 @@ public class ToolsGenerate extends BaseCommand {
 
             if (importToolset) {
                 System.err.print("\n\nImporting toolset...");
-                importToolset(toolReferences, host);
+                int failures = importToolset(toolReferences, host);
+                if (failures > 0) {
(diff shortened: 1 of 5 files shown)


apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/tools/ToolsGenerate.java
```diff
@@ -97,7 +97,11 @@ public Integer doCall(Terminal terminal, WanakuPrinter printer) {
 
             if (importToolset) {
                 System.err.print("\n\nImporting toolset...");
-                importToolset(toolReferences, host);
+                int failures = importToolset(toolReferences, host);
+                if (failures > 0) {
+                    System.err.printf("Import completed with %d failure(s)%n", failures);
```
orpiske: WHy the other uses printer.printErrorMessage ?

PR #1587, opened 2026-07-16
[#1518] Local mode should warn about disabled authentication

### LINKED ISSUE #1518: Local mode should warn about disabled authentication
## Source

Reported in blog post: [Wanaku: Getting Started](https://www.the-main-thread.com/p/wanaku-getting-started)

## Description

Local mode disables authentication automatically (via the `noauth` profile). The author notes: "Local mode disables authentication automatically, which is fine for a laptop demo and wrong for anything shared."

There is no warning or notice to the user that auth is disabled when starting in local mode, especially if the instance is exposed beyond localhost.

## Expected behavior

- On startup in local mode, print a visible warning that authentication is disabled
- If the server binds to a non-loopback address (e.g., `0.0.0.0`), the warning should be more prominent
- Consider documenting the security implications in the getting-started guide

## Acceptance Criteria

- Add a startup warning when running in local/noauth mode
- Make the warning more prominent when bound to a non-loopback interface

rnetuka:
Issue: https://github.com/wanaku-ai/wanaku/issues/1518

## Summary by Sourcery

New Features:
- Display a warning message indicating that authentication is disabled when running in local mode.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/start/StartLocal.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/commands/start/StartLocal.java
@@ -162,6 +162,7 @@ public class StartLocal extends StartBase {
                 return EXIT_OK;
             }
         }
+        printer.printWarningMessage("Authentication is disabled in local mode");
         startWanaku();
         return EXIT_OK;
     }


rnetuka:
@orpiske Is there any way how to obtain server bind address?

Perhaps:
```
@ConfigProperty(name = "wanaku.service.registration.uri")
String registrationUrl;

...

List<String> loopback = List.of("localhost", "127.0.0.1", "::1", "0.0.0.0");
if (registrationUrl == null || loopback.contains(registrationUrl)) {
    printer.printWarningMessage("Authentication is disabled in local mode");
} else {
    printer.printWarningMessage("WARNING: Authentication is disabled in local mode");
}
```
will work?

orpiske:
@rnetuka I think that you can try reading the quarkus.http.port property in the router

orpiske:
Oh, but you need to do that in the CLI, right? What would be the value of the `--host` flag, I think ... The `quarkus.http.port` can only be read from within the router ...

rnetuka:
This piece of code works for me:
```
String registrationUrl = ConfigProvider.getConfig()
                .getOptionalValue("wanaku.service.registration.uri", String.class)
                .orElse(null);
if (registrationUrl == null || List.of("localhost", "127.0.0.1", "::1", "0.0.0.0").contains(registrationUrl)) {
    printer.printWarningMessage("Authentication is disabled in local mode");
} else {
    printer.printWarningMessage("WARNING: Authentication is disabled in local mode!");
}
```
if I run it with `wanaku start local -Dwanaku.service.registration.uri=www.foo.bar:8080`. I'm not sure however if It pays up just to distinguish between two warning messages? Perhaps we could leave only the simple warning about auth being disabled. @orpiske WDYT?

orpiske:
@rnetuka Yeah, just a simple warning should be fine ... so the user is aware.

PR #1603, opened 2026-07-20
refactor(AuthCredentialStore): move default credentials file path to a dedicated method

### LINKED ISSUE #1592: CLI: AuthCredentialStore eagerly creates credentials directory, fails in native image with custom WANAKU_HOME
## Description

The `AuthCredentialStore` class eagerly creates the credentials directory in its constructor, which causes the CLI to fail with `Failed to create credentials directory` when `WANAKU_HOME` is set to a custom path.

## Root Cause

`AuthCredentialStore.java:26` defines:
```java
private static final String DEFAULT_CREDENTIALS_FILE = WanakuHome.get() + File.separator + "credentials";
```

This `static final` field evaluates `WanakuHome.get()` at class-initialization time. In the GraalVM native image, if this class is initialized at build time, the resolved path captures the **build machine's** home directory, not the runtime `WANAKU_HOME`. The constructor unconditionally calls `ensureCredentialsDirectoryExists()`, which fails when the build-time path doesn't exist on the target machine.

Note: `WanakuHome.defaultWanakuHome()` already uses a method (not a field) specifically to avoid this GraalVM pitfall, as noted in its Javadoc. The same pattern should be applied to `DEFAULT_CREDENTIALS_FILE`.

## Steps to Reproduce

1. Build the Wanaku CLI as a native image
2. Set `WANAKU_HOME` to a custom directory (e.g., `/tmp/wanaku-test`)
3. Run any CLI command (e.g., `wanaku tools list --host ...`)
4. Observe: `An error occurred: Failed to create credentials directory`

## Expected Behavior

The CLI should:
1. Resolve `DEFAULT_CREDENTIALS_FILE` at runtime (using a method instead of `static final`)
2. Respect the `WANAKU_HOME` environment variable for credential storage
3. Not fail on non-auth commands if the credentials directory can't be created

## Affected Version

0.2.0-SNAPSHOT (native CLI binary)

## Source References

- `apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/AuthCredentialStore.java:26` (static final field)
- `apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/AuthCredentialStore.java:47` (eager constructor call)
- `apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/AuthCredentialStore.java:266-280` (directory creation)
- `core/core-util/src/main/java/ai/wanaku/core/util/WanakuHome.java:30` (method pattern to follow)

matheusandre1:
Closes: #1592

## Summary by Sourcery

Enhancements:
- Introduce a getDefaultCredentialsFile() helper to encapsulate construction of the default credentials file path used by AuthCredentialStore.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/AuthCredentialStore.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/AuthCredentialStore.java
@@ -23,7 +23,6 @@ public class AuthCredentialStore {
 
     private static final Logger LOG = Logger.getLogger(AuthCredentialStore.class);
 
-    private static final String DEFAULT_CREDENTIALS_FILE = WanakuHome.get() + File.separator + "credentials";
     private static final String API_TOKEN_KEY = "api.token";
     private static final String REFRESH_TOKEN_KEY = "refresh.token";
     private static final String AUTH_MODE_KEY = "auth.mode";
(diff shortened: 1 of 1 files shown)


matheusandre1:
That was a bug I caused—my bad.

orpiske:
> That was a bug I caused—my bad.

No worries. Thanks for the fix!

PR #1616, opened 2026-07-20
Fix #1605: fix spurious null error in namespaces create without --name

orpiske:
## Summary

- Fix `NullPointerException` in `WanakuPrinter.printAsMap()` when map values contain null entries
- The key-filtering logic now handles null values gracefully instead of throwing

## Test plan

- [x] New `WanakuPrinterPlainModeTest` covering null value scenarios
- [x] New `NamespacesCreateTest` verifying the create-without-name flow

## Summary by Sourcery

Handle null values safely in CLI map printing and ensure namespace creation works when name is omitted.

Bug Fixes:
- Prevent NullPointerException in WanakuPrinter.printAsMap when maps contain null values.
- Ensure namespaces can be created without specifying a name in the CLI command.

Tests:
- Add WanakuPrinterPlainModeTest coverage for printing maps containing null values.
- Extend NamespacesCreateTest to cover create-without-name flow.

PR #1628, opened 2026-07-20
Fix #1605: handle null values in WanakuPrinter.printAsMap

### LINKED ISSUE #1605: cli: namespaces create without --name returns spurious null error despite succeeding
## Description

Running `wanaku namespaces create --path <path> --plain` without the `--name` flag (pre-allocated namespace) creates the namespace server-side successfully but the CLI exits with code 1 and prints "An error occurred: null".

## Steps to Reproduce

```bash
wanaku namespaces create --host http://localhost:8080 --path "cleanup-target" --plain
```

## Expected Behavior

Exit code 0 with the namespace ID printed (the namespace is created successfully on the server).

## Actual Behavior

The namespace IS created server-side (a UUID is returned), but the CLI then prints "An error occurred: null" and exits with code 1. This breaks any automation relying on exit codes.

## Impact

- Severity: Medium
- Affects scripted/automated namespace provisioning workflows
- Pre-allocated namespaces cannot be reliably created without error handling workarounds

## Environment

- Wanaku 0.3.0-SNAPSHOT (local build)
- Reproduced across two independent QE runs (2026-07-14 and 2026-07-20)

orpiske:
Fixes #1605

## Summary

`WanakuPrinter.printAsMap` used `Collectors.toMap()` to filter map entries by key, but that collector throws `NullPointerException` when any value is `null`. When creating a pre-allocated namespace without the `--name` flag, the `name` field is `null`, causing the NPE to propagate to the exception handler which prints "An error occurred: null" and exits with code 1 -- even though the namespace was created successfully server-side.

## Changes

- Replace the stream-based `Collectors.toMap()` with a plain `HashMap` loop that accepts null values
- Remove the now-unused `Collectors.toMap` static import
- Add a test case for `printAsMap` with null values

## Test plan

- [x] Unit tests pass (`WanakuPrinterPlainModeTest`, `NamespacesCreateTest`)
- [x] Integration tests: pre-existing failures on main, unrelated to this change

## Summary by Sourcery

Handle null map values in WanakuPrinter.printAsMap to prevent erroneous failures when printing filtered maps.

Bug Fixes:
- Prevent NullPointerException in printAsMap when input maps contain null values.

Tests:
- Add a unit test verifying printAsMap correctly handles maps with null values without throwing and still prints non-null entries.

### DIFF
--- a/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/WanakuPrinter.java
+++ b/apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/WanakuPrinter.java
@@ -26,7 +26,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 import static java.lang.Boolean.FALSE;
 import static java.lang.Boolean.TRUE;
-import static java.util.stream.Collectors.toMap;
 import static org.jline.console.Printer.TableRows.EVEN;
 
 /**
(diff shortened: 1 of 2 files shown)


orpiske:
Closing as duplicate of PR #1616 which has the same fix plus an additional NamespacesCreateTest. The improved assertDoesNotThrow pattern from this PR will be incorporated as a follow-up.

PR #1640, opened 2026-07-21
Fix #1637: validate namespace names to reject non-alphanumeric/dash characters

### LINKED ISSUE #1637: Router: validate namespace names to reject non-alphanumeric/dash characters
### Description

The router should validate namespace names at creation time to ensure they only contain alphanumeric characters and dashes. Currently, invalid characters (such as a leading `/`) are accepted during namespace creation but cause `IllegalStateException: Invalid server name` later when the namespace path is used as a gRPC server name during tool/resource registration.

### Current behavior

A namespace can be created with a path like `/my-namespace` (containing a leading `/`). The creation succeeds, but subsequent operations that use the namespace path as a gRPC server name fail with:

```
IllegalStateException: Invalid server name: /my-namespace
```

### Expected behavior

The router should reject namespace names and paths that contain characters other than alphanumeric characters (`a-z`, `A-Z`, `0-9`) and dashes (`-`) at creation time, returning a clear validation error (e.g., HTTP 400) instead of accepting the invalid value and failing later with a generic 500 error.

### Suggested validation

Namespace names and paths should match the pattern: `^[a-zA-Z0-9][a-zA-Z0-9-]*$`

This would catch invalid inputs early and provide a clear error message to the user.

orpiske:
Fixes #1637

## Summary

Adds input validation for namespace names and paths at creation, update, and allocation time in the router backend. Namespace names must now match the pattern `^[a-zA-Z0-9][a-zA-Z0-9-]*$` (alphanumeric characters and dashes, starting with an alphanumeric character).

## Changes

- **NamespacesBean**: Added `validateNamespaceName()` method with regex-based validation, called from `create()`, `update()`, and `alocateNamespace()` methods
- **IllegalArgumentExceptionMapper**: New exception mapper that converts `IllegalArgumentException` to HTTP 400 (Bad Request) responses with a clear error message, instead of letting them fall through to the generic 500 handler
- **Tests**: Added unit tests for validation in `NamespacesBeanTest` (7 new test cases) and REST endpoint tests in `AbstractNamespacesResourceTest` (2 new test cases)

## Test plan

- [x] Unit tests pass locally (389 tests, 0 failures)
- [x] Full reactor build succeeds
- [x] Integration tests show no new failures (pre-existing failures identical to previous runs)


orpiske:
Closing as duplicate of PR #1641 — same fix created by overlapping workflow runs.

PR #1761, opened 2026-07-30
operator: Enhance HTTPS support and documentation

### LINKED ISSUE #607: Operator should support HTTPs deployments
The operator defaults to HTTP when setting up the auth proxy. This is not adequate

claudio4j:
- Added ingress and tls configuration to the wanaku router CRD.
- Sync generated k8s yaml files.
- Reworked the quarkus-openshift dependencies in a maven profile, activated when using -Dopenshift.
- Added support to customize operator deployment environment variables from custom resource annotations.
- Enhanced the deploy/deploy-to-dev-env.sh script to works in minikube and openshift.
- Enhanced documentation and tests related to the above changes.

Fix https://github.com/wanaku-ai/wanaku/issues/607

## Summary by Sourcery

Enhance router exposure configuration to support HTTPS via Kubernetes Ingress and OpenShift Route, add annotation-driven env var overrides across operator-managed deployments, and improve dev deployment and build workflows for Kubernetes/OpenShift.

New Features:
- Introduce spec.exposure on WanakuRouter CRD to control external access via Ingress, Route, or none, including TLS options and ingress metadata.
- Enable HTTPS URLs end-to-end for router status, auth proxy, and deployment templates, including default TLS behavior for OpenShift Routes.
- Allow operator-managed Deployments (router, capabilities, Camel routes, code execution engine) to derive and override environment variables from CR annotations with the env.wanaku.ai/* prefix.

Enhancements:
- Refine router reconciliation with spec validation, OpenShift detection guardrails, and improved Route/Ingress matching semantics for TLS and ingressClass changes.
- Expand tests for router exposure, TLS handling, environment-variable injection, and CRD type conversions to cover new behavior.
- Update Helm chart values and RBAC to support configurable operator replicas and proper Camel code execution engine permissions and scope bindings.
- Improve dev deployment script to work on both Minikube and OpenShift, including automatic ingress host resolution, OIDC secret management, and shared Kubernetes manifests.
- Add Maven profiles to switch between Jib and OpenShift container-image extensions based on the -Dopenshift flag for operator and tool services.
- Adjust Keycloak and usage docs to simplify setup, reflect new Kubernetes manifests, and add troubleshooting guidance for ingress buffer issues.

Documentation:
- Document annotation-based environment variable injection for operator-managed CRDs and the expanded WanakuRouter exposure/TLS fields in operator docs.
- Update contributing and usage guides for unified Kubernetes manifests, new build profiles, and Minikube/OpenShift deployment flows.
- Extend troubleshooting documentation with guidance for Keycloak hostname configuration and nginx proxy buffer issues when using Wanaku with HTTPS.

Tests:
- Add unit tests for router spec validation, exposure/TLS enums, environment annotation processing, and updated Route/Ingress matching.
- Extend capability, Camel route, and code execution engine reconciler tests to verify annotation-based env var overrides and additions.
- Add ingress and route factory tests to validate TLS, ingressClassName, and annotations wiring for external access resources.


orpiske:
Thanks! Regarding this change: `Reworked the quarkus-openshift dependencies in a maven profile, activated when using -Dopenshift.` Wouldn't this make the release process more complicated?

apps/wanaku-operator/src/main/java/ai/wanaku/operator/util/RouterResourceFactory.java
```diff
@@ -172,11 +213,37 @@ public static Ingress makeRouterIngress(WanakuRouter resource, String host) {
...
+            merged.putAll(exposureSpec.getAnnotations());
+            ingress.getMetadata().setAnnotations(merged);
+        }
+        IngressTLS ingressTls = new IngressTLS();
+        ingressTls.setHosts(List.of(host));
+        if (exposureSpec.getTls() != null
+                && StringHelper.isNotEmpty(exposureSpec.getTls().getSecretName())) {
+            ingressTls.setSecretName(exposureSpec.getTls().getSecretName());
+        }
+        ingress.getSpec().setTls(List.of(ingressTls));
```
orpiske, line 244: **Bug (Blocking):** `IngressTLS` is always created and set on the Ingress spec, even when no TLS is configured. When a user sets `type: ingress` with a host but omits `tls`, the generated Ingress will contain a `tls:` section with only `hosts` and null `secretName`. Most ingress controllers (nginx, traefik) interpret any TLS entry as requiring HTTPS termination, which will cause certificate errors or 503s for users who want plain HTTP.

The test `buildIngressWithoutTlsHasNoTlsStanza` passes only because it uses a router with null `exposure` entirely — it doesn't test the realistic `type: ingress, host: X, no tls` scenario.

Suggested fix — only create the TLS stanza when TLS is actually configured:

```suggestion
        if (exposureSpec.getTls() != null
                && StringHelper.isNotEmpty(exposureSpec.getTls().getSecretName())) {
            IngressTLS ingressTls = new IngressTLS();
            ingressTls.setHosts(List.of(host));
            ingressTls.setSecretName(exposureSpec.getTls().getSecretName());
            ingress.getSpec().setTls(List.of(ingressTls));
        }
```

And consider adding a test for `type: ingress` with host but no `tls`.

apps/wanaku-operator/src/main/java/ai/wanaku/operator/wanaku/WanakuRouterReconciler.java
```diff
@@ -103,88 +106,98 @@ public UpdateControl<WanakuRouter> reconcile(WanakuRouter resource, Context<Wana
...
-            host = createIngressAndGetHost(resource, namespace);
-        }
-        wanakuStatus.setHost("%s://%s".formatted(SCHEME, host));
-        wanakuStatus.setSseEndpoint("%s://%s/mcp/sse".formatted(SCHEME, host));
-        wanakuStatus.setStreamableEndpoint("%s://%s/mcp/".formatted(SCHEME, host));
+        String host = reconcileExternalAccess(resource, namespace);
+
+        wanakuStatus.setHost("https://" + host);
+        wanakuStatus.setSseEndpoint("https://%s/mcp/sse".formatted(host));
+        wanakuStatus.setStreamableEndpoint("https://%s/mcp/".formatted(host));
```
orpiske, line 164: **Bug (Blocking):** All status URLs are now hardcoded to `https://`, but when `exposure.type` is `NONE` (or null), `reconcileExternalAccess()` returns `"internal-" + name` — a ClusterIP hostname that serves plain HTTP on port 8080, not HTTPS. These status URLs are misleading and unusable.

The same issue propagates to `authProxy` when set to `"auto"` — it becomes `"https://internal-<name>"`, an unreachable HTTPS endpoint.

The previous code correctly used `http://` (via the `SCHEME` constant). Consider using the scheme based on exposure type:

```java
String scheme = (exposureSpec != null
        && exposureSpec.getType() != null
        && exposureSpec.getType() != WanakuTypes.ExposureType.NONE)
        ? "https" : "http";
wanakuStatus.setHost("%s://%s".formatted(scheme, host));
wanakuStatus.setSseEndpoint("%s://%s/mcp/sse".formatted(scheme, host));
wanakuStatus.setStreamableEndpoint("%s://%s/mcp/".formatted(scheme, host));
```

apps/wanaku-operator/src/main/java/ai/wanaku/operator/wanaku/WanakuTypes.java
```diff
@@ -19,6 +22,70 @@ private WanakuTypes() {}
...
+        NONE;
+
+        @JsonCreator
+        public static ExposureType fromValue(String v) {
+            if (v == null) return null;
+            for (ExposureType t : values()) {
+                if (t.name().equalsIgnoreCase(v)) return t;
+            }
+            throw new IllegalArgumentException(
+                    "Invalid spec.ingress.type '%s'. Valid values: Route, Ingress, None".formatted(v));
```
orpiske: **Non-blocking:** The error message still references `spec.ingress.type` but the field was renamed to `spec.exposure.type` in this PR. This will confuse users who see this error.

```suggestion
                    "Invalid spec.exposure.type '%s'. Valid values: Route, Ingress, None".formatted(v));
```

apps/wanaku-operator/src/main/java/ai/wanaku/operator/util/RouterResourceFactory.java
```diff
@@ -135,11 +141,47 @@ public static Route makeRouterExternalService(WanakuRouter resource) {
...
+            return;
+        }
+        TLSConfig tlsConfig = new TLSConfig();
+        if (ingressSpec.getTls() == null) {
+            tlsConfig.setTermination(TlsTermination.EDGE.toValue());
+            tlsConfig.setInsecureEdgeTerminationPolicy(InsecureEdgeTerminationPolicy.REDIRECT.toValue());
+
+        } else {
+            WanakuTypes.TlsSpec tlsSpec = ingressSpec.getTls();
+            if (tlsSpec.getTermination() == null) {
```
orpiske, line 161: **Design concern (Non-blocking):** When `spec.exposure.tls` is null, the operator defaults to edge termination + insecure redirect (safe default). But when `tls` is *present* with `termination: null`, this method returns early — no TLS at all.

This creates a trap: a user adding `tls: {}` to configure only `insecureEdgeTerminationPolicy` would accidentally disable TLS entirely. Consider defaulting `termination` to `edge` when `tls` is present but `termination` is null, consistent with the null-tls default behavior.

apps/wanaku-operator/src/main/java/ai/wanaku/operator/wanaku/WanakuRouterSpec.java
```diff
@@ -24,12 +24,12 @@ public void setImagePullPolicy(String imagePullPolicy) {
...
 
-    public WanakuTypes.IngressSpec getIngress() {
-        return ingress;
+    public WanakuTypes.ExposureSpec getExposure() {
+        return exposure;
     }
 
-    public void setIngress(WanakuTypes.IngressSpec ingress) {
-        this.ingress = ingress;
+    public void setExposure(WanakuTypes.ExposureSpec ingress) {
```
orpiske: **Nit:** The parameter name is still `ingress` from before the rename.

```suggestion
    public void setExposure(WanakuTypes.ExposureSpec exposure) {
```

apps/wanaku-operator/src/main/java/ai/wanaku/operator/util/EnvironmentVariableHelper.java
```diff
@@ -16,8 +18,55 @@
...
+     * @return a mutable list of {@link EnvVar} objects; never null
+     */
+    public static List<EnvVar> extractAnnotationEnvVars(Map<String, String> annotations) {
+        List<EnvVar> result = new ArrayList<>();
+        if (annotations == null || annotations.isEmpty()) {
+            return result;
+        }
+        for (Map.Entry<String, String> entry : annotations.entrySet()) {
+            if (entry.getKey().startsWith(ANNOTATION_ENV_PREFIX)) {
+                String name = entry.getKey().substring(ANNOTATION_ENV_PREFIX.length());
```
orpiske, line 51: **Suggestion:** No validation that the post-prefix name is non-empty. An annotation key `env.wanaku.ai/` (no suffix) would produce an env var with empty name `""`, which Kubernetes will reject at pod scheduling time with an opaque error.

Consider adding:
```java
String name = entry.getKey().substring(ANNOTATION_ENV_PREFIX.length());
if (name.isEmpty()) {
    continue;
}
```

orpiske (changes requested):
Thank you for this substantial PR, Claudio! The `spec.exposure` CRD redesign is clean and extensible, the annotation-based env var injection is well-architected with consistent override-last semantics across all four reconcilers, and the test coverage is excellent. Great work overall.

There are a couple of correctness issues to address before merge, plus the CI static-analysis check is failing (3 PMD violations in `wanaku-operator`).

**Summary of findings:**

| # | Severity | Finding |
|---|---|---|
| 1 | Blocking | `applyIngressExtras()` unconditionally creates IngressTLS even when no TLS is configured |
| 2 | Blocking | Hardcoded `https://` in status URLs when `exposure.type` is `NONE` |
| 3 | Blocking | CI static-analysis failing (3 PMD violations) |
| 4 | Non-blocking | Stale error message in `ExposureType.fromValue()` — says `spec.ingress.type` |
| 5 | Non-blocking | `imagePullPolicy` changed from `IfNotPresent` to `Always` in templates — intentional? |
| 6 | Non-blocking | TLS defaults asymmetry: `tls: {}` (present but empty) disables TLS instead of defaulting to edge |
| 7 | Suggestion | Parameter name `ingress` in `setExposure()` setter |
| 8 | Suggestion | Missing validation for empty env var names from annotations |
| 9 | Suggestion | `echo $out` unquoted in deploy script |
| 10 | Nit | Doc typos: "Openshift" → "OpenShift", "shoud" → "should", "URLS" → "URLs" |

See inline comments for details on each finding.

> **Note:** This review focuses on correctness, project conventions, and architectural concerns. It does not replace specialized tools like SonarCloud, CodeRabbit, or Sourcery.

apps/wanaku-operator/src/main/java/ai/wanaku/operator/util/RouterResourceFactory.java
```diff
@@ -172,11 +213,37 @@ public static Ingress makeRouterIngress(WanakuRouter resource, String host) {
                 .getService()
                 .setName("internal-" + deploymentName);
 
+        applyIngressExtras(ingress, resource.getSpec().getExposure(), host);
         ingress.addOwnerReference(resource);
 
         return ingress;
     }
 
+    private static void applyIngressExtras(Ingress ingress, WanakuTypes.ExposureSpec exposureSpec, String host) {
```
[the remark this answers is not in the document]
claudio4j: https is the default when type is ingress or route.

apps/wanaku-operator/src/main/java/ai/wanaku/operator/util/EnvironmentVariableHelper.java
```diff
@@ -16,8 +18,55 @@
  */
 public final class EnvironmentVariableHelper {
 
+    static final String ANNOTATION_ENV_PREFIX = "env.wanaku.ai/";
```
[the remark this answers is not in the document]
claudio4j: fixed.

claudio4j:
> Regarding this change: Reworked the quarkus-openshift dependencies in a maven profile, activated when using -Dopenshift. Wouldn't this make the release process more complicated?

I think it doesn't affect the release process, since the container image build will work as before, with no additional parameters.

Both dependencies (jib and openshift) can be set with no profiles, but when building the container images, a parameter is required to let the quarkus container image builder know which one to use, for example: `-Dquarkus.container-image.builder=jib`.
To summarize, 
- use jib builder: no parameters, as default is `jib`
- openshift: -Dquarkus.container-image.builder=openshift

For either case, just for openshift it's going to require the additional parameter, so the `-Dopenshift` is easier to type.

apps/wanaku-operator/src/main/java/ai/wanaku/operator/util/RouterResourceFactory.java
claudio4j: It's intended to  always create the tls and expose it securely with https.

apps/wanaku-operator/src/main/java/ai/wanaku/operator/wanaku/WanakuRouterReconciler.java
claudio4j: Do you think there is an use case to deploy the mcp router in a kubernetes cluster, but not exposing it ? Probably we should remove the `none` option, wdyt ?

orpiske: > Do you think there is an use case to deploy the mcp router in a kubernetes cluster, but not exposing it ? 

I don't think so ...

orpiske:
> I think it doesn't affect the release process, since the container image build will work as before, with no additional parameters.

AHHH, got it ... thanks! I think it should fine, then

orpiske (approved):
Thank you for addressing the previous review feedback, Claudio! The new commit fixes the items cleanly:

**Resolved:**
- Error message updated from `spec.ingress.type` to `spec.exposure.type`
- Parameter name `ingress` → `exposure` in `setExposure()` setter
- Env var name validation added with `ENV_VAR_NAME_PATTERN` regex and thorough tests (empty suffix, hyphens, digit-leading names)
- All 3 PMD violations fixed — CI is green
- Documentation typos fixed ("OpenShift", "should", "Kubernetes", "URLs")

**Design decisions accepted:**
- IngressTLS always created for HTTPS-by-default — understood and reasonable for a security-first approach.
- Hardcoded `https://` in status URLs — this makes sense given the discussion about removing `NONE`. See remaining note below.

**Remaining non-blocking notes (for follow-up):**

1. **`ExposureType.NONE` resolution** — Per the discussion, if there's no use case for deploying without external access, removing `NONE` would clean up the code path. As long as `NONE` exists, `reconcileExternalAccess()` returns `internal-<name>` and the status URLs report `https://internal-<name>` (an HTTP-only ClusterIP), and `authProxy: auto` becomes `https://internal-<name>` (unreachable). If `NONE` is kept, the scheme should be conditional.

2. **TLS defaults asymmetry in `applyRouteTls()`** — When `spec.exposure.tls` is null → defaults to edge + redirect (good). When `tls` is present but `termination` is null → returns early, no TLS at all. A user adding `tls: {}` to set only `insecureEdgeTerminationPolicy` would accidentally disable TLS. Consider defaulting `termination` to `edge` when `tls` is present.

Overall this is a well-engineered PR with excellent test coverage. Approving — the remaining items can be addressed in a follow-up.

> **Note:** This review focuses on correctness, project conventions, and architectural concerns. It does not replace specialized tools like SonarCloud, CodeRabbit, or Sourcery.

PR #1763, opened 2026-08-01
fix: label filtering to capabilities API and update related components

### LINKED ISSUE #1554: capabilities list -e label expression filter does not filter results
## Description

`wanaku capabilities list -e "nonexistent=true"` returns all capabilities instead of filtering by the label expression. The same `-e` flag works correctly on `tools list`.

## Steps to Reproduce

1. Start the local stack with the HTTP tool capability (which has no labels)
2. Run: `wanaku capabilities list --host http://localhost:8080 -e "nonexistent=true" --plain`
3. Observe: the HTTP capability is returned despite not matching the filter

## Expected

Empty result set (no capabilities have label `nonexistent=true`).

## Root Cause

In `CapabilitiesHelper.java`, the `fetchAndMergeCapabilities(CapabilitiesService, String labelFilter)` method accepts the `labelFilter` parameter but never uses it. The filter is passed from `CapabilitiesList.doCall()` but discarded in the helper — it calls `fetchCapabilities(capabilitiesService)` without forwarding the expression.

**File:** `apps/wanaku-cli/src/main/java/ai/wanaku/cli/main/support/CapabilitiesHelper.java`, lines 85-96.

## Found by

QE run 2026-07-14, test plan `cli-capabilities-management`, test 2.4.

matheusandre1:
Fixes: #1554

## Summary by Sourcery

Add label-based filtering support to the capabilities API and propagate the new query parameter through backend, CLI, and UI clients.

New Features:
- Introduce optional labelFilter query parameter on the capabilities listing endpoint to filter returned capabilities.
- Expose label-filtered capabilities listing through the CapabilitiesService interface and CLI helper.
- Support passing capabilities query parameters, including labelFilter, from the admin UI via the generated API client and models.

Bug Fixes:
- Ensure invalid label filter expressions on the capabilities endpoint return a structured error instead of silently succeeding.

Enhancements:
- Implement label expression evaluation for capabilities using a shared LabelExpressionParser, with defined behavior when capabilities have no labels.
- Simplify tools and resources listing to no longer apply legacy label-based filtering logic.

Documentation:
- Update OpenAPI specification to document the labelFilter query parameter on the capabilities endpoint.

Tests:
- Add backend tests covering capabilities label filtering behavior, blank filters, and error handling for invalid label expressions.


orpiske:
Thanks @matheusandre1. 

This has become deprecated with the move of the routing engine to Praxis. As the Praxis-based engine doesn't have the same persistence layer, I don't think it can be applied in this case. We may still apply some of the ideas on top of the peersistence layer on [wanaku-barn](https://github.com/wanaku-ai/wanaku-barn) project.

PR #1858, opened 2026-08-20
fix #1851: hoist parse_body to avoid double parsing in tool_call filter

### LINKED ISSUE #1851: Fix double parse_body call in tool_call filter
## Problem

In `filters/src/tool_call.rs`, `parse_body(body)` is called twice:

- Line 67-68: called in the error path when `mcp.name` metadata is missing
- Line 76: called again in the main path

The body is parsed and deserialized redundantly.

## Suggested Fix

Hoist the `parse_body` call to the top of `handle_body`, before the metadata check:

```rust
async fn handle_body(&self, ctx: &mut HttpFilterContext<'_>, body: &mut Option<Bytes>) -> Result<FilterAction, FilterError> {
    let body_ref = body.as_ref().ok_or(/* ... */)?;
    let parsed = parse_body(body_ref);  // parse once here

    let tool_name = match ctx.get_metadata(MCP_NAME_KEY) {
        Some(name) => name.to_string(),
        None => {
            // use parsed.id from above instead of re-parsing
            // ...
        }
    };
    // use parsed for the rest of the function
}
```

## Files to Change

- `filters/src/tool_call.rs` — hoist `parse_body` call to top of `handle_body`

## Impact

Minor performance fix — eliminates redundant JSON deserialization per tool call request.

orpiske:
## Summary

- Hoists `parse_body(body)` call to the top of `handle_body` in `ToolCallFilter`, so the body is parsed once instead of twice (previously parsed in both the error path at line 67 and the main path at line 76)
- No behavior change — same parse result, just computed earlier

## Test plan

- [x] `cargo test -p wanaku-filters` — all 60 tests pass
- [ ] CI (clippy + tests) via PR Build workflow

## Summary by Sourcery

Parse tool-call request bodies once to eliminate redundant work while preserving existing behavior.

Bug Fixes:
- Avoid parsing tool-call request bodies twice when handling missing tool-name errors.

Enhancements:
- Hoist request-body parsing so the parsed result is shared across all handling paths without changing behavior.

### DIFF
--- a/filters/src/tool_call.rs
+++ b/filters/src/tool_call.rs
@@ -61,10 +61,11 @@ impl ToolCallFilter {
             return Ok(FilterAction::Continue);
         }
 
+        let mut parsed = parse_body(body);
+
         let tool_name = match ctx.get_metadata(crate::MCP_NAME_KEY) {
             Some(n) => n.to_owned(),
             None => {
-                let parsed = parse_body(body);
                 return Ok(crate::response::json_rpc_error(&parsed.id, crate::response::JSONRPC_INVALID_PARAMS, "missing tool name in tools/call"));
             }
         };
@@ -72,8 +73,6 @@ impl ToolCallFilter {
         let namespace = ctx
             .get_metadata(crate::namespace::NAMESPACE_METADATA_KEY)
             .unwrap_or(wanaku_apis::registry::DEFAULT_NAMESPACE);
-
-        let mut parsed = parse_body(body);
 
         let conversation_id = parsed.arguments
             .remove(wanaku_apis::correlation::REQUEST_ID_ARG)


orpiske:
**Integration test note:** Ran `Integration Tests` at [orpiske/wanaku-tests](https://github.com/orpiske/wanaku-tests/actions/runs/32403546980) with `version=0.3.0-SNAPSHOT`. The workflow failed with a 404 at the "Download artifacts" step — the download script expects Java-era artifact names (`wanaku-router-backend-*.zip`) but the early-access release now contains Rust binaries (`wanaku-praxis-*.tar.gz`). This is a pre-existing infrastructure mismatch unrelated to this PR's changes.

PR Build CI (clippy + cargo test + UI E2E) passed.

orpiske:
**Integration test update (Full Integration Test From Source):** All actual tests passed. The run was marked as failure due to a pre-existing skip-threshold guard: `Skip threshold exceeded: 33% of tests were skipped (1/3), threshold is 30%` in router-tests. This affects all branches equally and is unrelated to this PR.

PR #1863, opened 2026-08-20
fix #1854: move env var read from InterceptFeature::new() to load_env_config()

### LINKED ISSUE #1854: Fix inconsistent config loading in InterceptFeature
## Problem

Feature config loading patterns are inconsistent across features:

- `evaluator` and `plugins` use `load_yaml_config` with serde deserialization
- `chat`, `intercept`, `mcp-metadata` have no-op `load_yaml_config`
- `InterceptFeature` reads `WANAKU_INTERACTION_CAPACITY` in its `new()` constructor instead of in `load_env_config()`

The `Feature` trait provides `load_env_config()` specifically for env var configuration, but `InterceptFeature` bypasses it.

## Suggested Fix

Move the `WANAKU_INTERACTION_CAPACITY` env var read from `InterceptFeature::new()` to `InterceptFeature::load_env_config()`:

```rust
fn load_env_config(&mut self) {
    if let Ok(val) = std::env::var("WANAKU_INTERACTION_CAPACITY") {
        if let Ok(cap) = val.parse::<usize>() {
            self.capacity = cap;
        }
    }
}
```

Adjust `new()` to use a default capacity, overridden when `load_env_config()` is called during startup.

## Files to Change

- `features/intercept/src/lib.rs` — move env var read from `new()` to `load_env_config()`

## Impact

Consistency — all features follow the same config loading lifecycle.

orpiske:
## Summary

- Moves `WANAKU_INTERACTION_CAPACITY` env var reading from `InterceptFeature::new()` to `load_env_config()`, consistent with the Feature trait lifecycle
- Wraps `store` field in `RwLock<InMemoryInteractionStore>` since `load_env_config(&self)` takes a shared reference
- Server lifecycle guarantees `load_env_config()` runs before `pipeline_extensions()` and `handle_route()`, so there's no concurrent write contention at runtime
- Poisoned lock handling follows codebase patterns: `pipeline_extensions` falls back to default-capacity store, `handle_route` returns `None`
- `InMemoryInteractionStore` clone is cheap (Arc clone)

## Test plan

- [x] `cargo build` — full project compiles
- [x] `cargo clippy -p wanaku-feature-intercept` — no warnings
- [ ] CI (clippy + tests) via PR Build workflow

## Summary by Sourcery

Move interaction store environment configuration into the feature configuration phase while preserving reliable request handling.

Bug Fixes:
- Apply the interaction-capacity environment setting during feature configuration rather than construction, aligning it with the feature lifecycle.

Enhancements:
- Make the interaction store safely replaceable and readable through the feature lifecycle while handling poisoned locks gracefully.

### DIFF
--- a/features/intercept/src/lib.rs
+++ b/features/intercept/src/lib.rs
@@ -80,7 +80,10 @@ impl Feature for InterceptFeature {
             Ok(guard) => guard.clone(),
             Err(e) => {
                 tracing::warn!(error = %e, "interaction store lock poisoned");
-                return None;
+                return Some(wanaku_apis::http_response::json_err(
+                    http::StatusCode::INTERNAL_SERVER_ERROR,
+                    "interaction store unavailable",
+                ));
             }
         };
         Some(match route {
--- a/features/intercept/src/lib.rs
+++ b/features/intercept/src/lib.rs
@@ -3,6 +3,8 @@
 pub mod filter;
 mod routes;
 
+use std::sync::RwLock;
+
 use http::Response;
 use praxis_filter::{FilterRegistry, PipelineExtension, RequestExtensions};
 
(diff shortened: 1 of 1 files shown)


orpiske:
**Integration test note:** Ran `Integration Tests` at [orpiske/wanaku-tests](https://github.com/orpiske/wanaku-tests/actions/runs/32403546980) with `version=0.3.0-SNAPSHOT`. The workflow failed with a 404 at the "Download artifacts" step — the download script expects Java-era artifact names (`wanaku-router-backend-*.zip`) but the early-access release now contains Rust binaries (`wanaku-praxis-*.tar.gz`). This is a pre-existing infrastructure mismatch unrelated to this PR's changes.

PR Build CI (clippy + cargo test + UI E2E) passed.

orpiske:
**Integration test update (Full Integration Test From Source):** All actual tests passed. The run was marked as failure due to a pre-existing skip-threshold guard: `Skip threshold exceeded: 33% of tests were skipped (1/3), threshold is 30%` in router-tests. This affects all branches equally and is unrelated to this PR.

PR #1881, opened 2026-08-21
Fixed LLM Chat UI and backend for use with Praxis

rnetuka:
Here's a fix for both LLM chat UI and backend, working together with Praxis

- Selected LLM is configured in Praxis, it is not displayed anymore
- Listing models doesn't take any parameter (because of above)
- API Key is no longer entered from UI

- I've added a check for namespace used for filtering tools on UI side. In my case, I've stored the public namespace which  doesn't exists anymore and it broke my UI. I think it's good to check if the value returned from config is valid anymore before use
- On backend, the v1 is already part of the base URL. Adding it again results in duplicates, for example /api/v1/v1/chat/completions

## Summary by Sourcery

Align the LLM chat interface and backend integration with Praxis-managed inference and namespace-based MCP tools.

New Features:
- Enable LLM chat tool calling through MCP, including execution of requested tools and display of tool requests and responses.
- Use Praxis-managed model and API configuration with a single selected model instead of provider-specific UI configuration.

Bug Fixes:
- Prevent duplicate API version segments when sending inference requests.
- Validate the configured tool namespace and fall back to the default namespace when it is no longer available.

Enhancements:
- Simplify model discovery and chat configuration around the Praxis-provided inference setup.
- Preserve tool-call conversation history and support iterative responses after tool execution.
- Switch MCP connections to the streamable HTTP transport and use the selected namespace.

Chores:
- Remove the LLM provider selection UI and avoid persisting the API key in local storage.


features/chat/src/routes.rs
```diff
@@ -118,16 +106,6 @@ pub(crate) async fn handle_chat_completions(
...
-    let request_api_key = request
-        .get("apiKey")
-        .and_then(serde_json::Value::as_str)
-        .unwrap_or("");
-    let effective_key = if request_api_key.is_empty() {
-        api_key
-    } else {
-        request_api_key
-    };
-
```
orpiske: I need to think about this part in particular. There may be some cases where organizations will want to use their own API keys (the one set on the server). 

I think this can also break if the inference endpoint is configured to load balance between different models from different providers.

orpiske (commented):
In general, it looks good to me. I just need to think about the effective API key use case.

orpiske (approved):
LGTM, thanks!

Issue #1875, opened 2026-08-21
Add side-effect-free policy validation, simulation, and traffic replay


orpiske:
## Description

Add tooling to validate and simulate governance policy changes before they become active.

Operators should be able to submit a candidate policy, evaluate synthetic or previously observed actions against it, compare the results with the active policy, and receive a structured impact report.

Simulation must not forward requests, resolve credentials, modify registry state, activate configuration, or produce other external side effects.

## Current Gap

Evaluator updates submitted through `PUT /api/v1/evaluators` are loaded directly into the active runtime configuration.

Wanaku does not currently provide a way to answer questions such as:

- Is this policy structurally and operationally valid?
- Which actions would change from allowed to denied?
- Which actions would change from denied to allowed?
- Which rule or evaluator would produce each decision?
- Does a rule never match any known capability?
- Does a candidate policy rely on unavailable processors or schemas?
- Would a recommendation broaden access beyond its supporting evidence?
- What would happen if the policy were applied to recent governed traffic?
- Can the policy be checked in CI without activating it?

This makes policy review dependent on manual inspection and increases the risk of outages, unintended access, and configuration changes whose impact is only discovered after activation.

## Proposed Capabilities

Provide three related but distinct operations:

1. Validation
2. Single-action simulation
3. Batch replay and comparison

All operations should accept either:

- An existing policy revision
- An inline candidate configuration
- A structured policy patch applied to a specified base revision

Inline candidates and patches must remain transient and must not become active implicitly.

## Validation

Validation should perform every deterministic check available without processing traffic.

At minimum, validate:

- Configuration syntax and schema
- Unique evaluator and rule identifiers
- Trigger and selector syntax
- JSON Pointer predicates
- Referenced namespaces
- Referenced registry labels
- Referenced WASM processors
- WASM component compatibility
- Result-schema compilation
- Fine-grained policy semantics from #1870
- Enforcement posture from #1872
- Credential-binding references from #1874 without resolving credentials
- Unsupported or contradictory configuration
- Bounded resource and configuration-size limits

Validation should report all safely discoverable errors in one response rather than stopping at the first error.

Warnings should be distinguished from errors.

Examples of warnings include:

- A rule that matches no currently registered capability
- A rule fully shadowed by an explicit deny
- A broad wildcard selector
- A permissive no-match baseline
- A fail-open configuration
- A disabled protected scope
- A reference to a capability that is currently unavailable

Validation must not contact LLM providers or upstream MCP servers.

[...]

Issue #1872, opened 2026-08-21
Add explicit enforcement modes and fail-safe policy baselines


orpiske:
## Description

Add explicit governance enforcement modes and fail-safe behavior for unmatched, invalid, or unavailable policies.

Operators should be able to determine whether a policy is actively enforcing decisions, observing traffic without enforcement, or intentionally disabled. They should also be able to configure what happens when Wanaku cannot reach a governance decision.

The effective enforcement posture must be visible, deterministic, and covered by audit events and metrics.

## Current Gap

Evaluator configuration currently supports an `on_error` value of `continue` or `block`, with `continue` as the default.

This leaves several aspects of the effective governance posture implicit:

- There is no explicit distinction between enforcement and audit-only evaluation.
- An absent or unmatched evaluator normally allows processing to continue.
- Evaluator failures default to continuing the request.
- Startup configuration errors can leave expected controls unavailable.
- WASM compilation failures are logged but do not expose a policy readiness state.
- There is no common behavior for LLM failures, invalid results, missing processors, WASM traps, or internal evaluator failures.
- Operators cannot query whether a namespace is protected, unconfigured, invalid, or degraded.
- It is difficult to determine whether observed decisions were enforced or merely recorded.

For a governance proxy, silently proceeding when a required control is unavailable can create a gap between configured intent and actual enforcement.

## Proposed Behavior

Introduce an explicit governance posture with three independent concepts:

1. Enforcement mode
2. No-match baseline
3. Failure behavior

Keeping these concepts separate avoids overloading a single `on_error` option with unrelated semantics.

## Enforcement Modes

Support the following modes:

### `enforce`

Policy decisions affect request processing.

- Allow decisions continue through the remaining governance pipeline.
- Deny decisions block the action.
- Response mutations and filtering actions are applied.
- Failures follow the configured failure behavior.

### `audit`

Policies and evaluators run, but their decisions do not block or mutate the request or response.

- Would-be decisions are recorded in audit events and metrics.
- Registry writes and other evaluator side effects are suppressed.
- WASM guests must receive read-only host capabilities where necessary.
- Failures are recorded but do not block traffic.

Audit mode must not be implemented by merely converting a final block into a continue after side effects have already occurred.

### `disabled`

The configured policy scope is intentionally bypassed.

- Evaluation is not performed.
- The disabled state is visible through status APIs and metrics.
- A structured warning is emitted when a protected scope is disabled.
[...]

Issue #1869, opened 2026-08-21
Add a structured governance decision audit trail with redaction and persistence


orpiske:
## Description

Add a structured audit trail for governance decisions made while proxying MCP, agent, and inference traffic.

The audit trail should provide durable evidence of who requested an action, what was requested, which governance controls evaluated it, what decision was made, and what happened afterward. Audit records must use a stable schema, support correlation across the request pipeline, and avoid exposing credentials or sensitive payload data.

## Current Gap

Wanaku currently provides two related but distinct observability mechanisms:

- The intercept feature stores complete request and response bodies in a bounded in-memory interaction store.
- The metrics feature records aggregate counters and durations for filters and evaluators.

These mechanisms are useful for conversation context and operational monitoring, but they do not provide a durable governance audit trail.

The current implementation does not provide:

- A stable event schema for governance decisions
- Persistent decision history
- A unique event identifier
- End-to-end request and decision correlation
- Actor or workload identity fields
- Policy or evaluator revision attribution
- Structured reason codes for allow, block, warn, or malformed outcomes
- Consistent redaction before persistence or export
- Querying by namespace, action, decision, actor, or time
- Configurable retention and export behavior

Interaction records also contain complete request and response bodies. Treating these records as audit events would risk retaining credentials, prompts, tool arguments, personal data, or backend responses unnecessarily.

## Proposed Behavior

Introduce a dedicated audit event model and audit store, separate from conversation interactions and aggregate metrics.

Wanaku should emit an audit event for governance-relevant outcomes, including:

- Accepted and rejected MCP operations
- Evaluator decisions
- Tool, resource, and prompt routing outcomes
- Inference routing and policy outcomes
- Management API changes to governance configuration
- Policy activation and rollback outcomes
- Credential resolution or injection outcomes without credential values
- Internal failures that affect enforcement

Each audit event should include, where applicable:

- Event ID
- Event schema version
- Timestamp
- Request and correlation IDs
- Conversation ID
- Authenticated actor, agent, or workload identity
- Namespace and audience
- Protocol and operation
- Target type and normalized target identifier
- Decision, such as `allow`, `block`, `warn`, `reject_malformed`, or `error`
- Stable reason code and human-readable explanation
- Evaluating filter or evaluator
- Policy or evaluator revision
- Upstream identifier
- Response status
- Duration
- Redaction metadata
- Additional structured attributes with bounded size

[...]

Issue #1331, opened 2026-06-14
Add object-level (per-tenant) authorization to the REST API and MCP namespaces


oscerd:
### Summary

There is no per-user or per-tenant (object-level) authorization in the router. Once authentication
is enforced, any authenticated principal can still read/modify/delete any other principal's objects
through the id/name endpoints, and the `ns-0..9` MCP namespaces enforce authentication but not
per-namespace authorization.

### Remediation

Introduce an owner/tenant attribute on stored entities and enforce it against the caller's
`SecurityIdentity`/token claim on every id/name lookup, and bind each MCP namespace tenant to its
authorized principal(s). (Alternatively, enable and configure the Keycloak policy enforcer.)

### Notes

Cross-cutting; best done after #1317-style authentication is in place (depends on the default-deny
work).

oscerd:
## Design note: object-level (per-tenant) authorization

Posting a design proposal here per discussion; this issue is **design-first**, not a single PR. It
assumes #1330 (default-deny of `/api/v1/*`) lands first — there is no point enforcing per-tenant
rules while the surface is anonymous.

### Problem

Authentication is now enforced, but **authorization is all-or-nothing**: any authenticated principal
can read/modify/delete any other principal's objects. Concretely:

- Every id/name lookup (`ToolsResource`, `ResourcesResource`, `PromptsResource`,
  `NamespacesResource`, `DataStoresResource`, `ServiceCatalogResource`, `ToolsetReposBean`, …)
  fetches/deletes purely by identifier with no ownership filter.
- The `ns-0..9` MCP namespaces enforce *authentication* (a valid token) but not *authorization* —
  a token minted for one namespace can call another namespace's tools.
- Stored entities (`DataStore`, `ToolReference`, `ResourceReference`, …) carry no owner/tenant
  attribute, so there is nothing to enforce against today.

### Proposed model

1. **Tenancy unit.** Recommend a **tenant/owner claim** (not raw `sub`) so the model supports both
   individual users and shared teams. Source it from a configurable JWT claim
   (`wanaku.authz.tenant-claim`, default e.g. `groups` or a custom `wanaku_tenant`). Fall back to
   `sub` when the claim is absent. A configurable `wanaku.authz.admin-role` bypasses the filter
   (operators/superusers).

2. **Persist the owner.** Add an `owner` (tenant id) field to the stored entities (via the
[...]

PR #1907, opened 2026-08-27
Fix #1870: add deterministic action policies

### LINKED ISSUE #1870: Add fine-grained action policies with deterministic deny precedence
## Description

Add declarative, fine-grained policies for governing individual MCP actions based on their target and structured request data.

Policies should be able to distinguish between operations such as calling a specific tool, reading a particular resource, or retrieving a prompt. When multiple applicable rules conflict, an explicit deny must take precedence over an allow.

This provides a deterministic enforcement layer for rules that do not require LLM evaluation while preserving the existing evaluator pipeline for contextual or semantic decisions.

## Current Gap

Evaluator triggers currently match only:

- MCP method
- Optional namespace

The evaluator state returns the first matching evaluator. It does not evaluate a set of applicable authorization rules or resolve conflicts between them.

As a result, Wanaku cannot directly express policies such as:

- Allow one tool while denying another in the same namespace
- Deny a sensitive operation exposed by an otherwise permitted tool
- Restrict a tool according to specific argument values
- Restrict resource access by URI
- Apply policy to tools or resources carrying particular labels
- Explain which rule caused an action to be denied
- Guarantee that a broad allow cannot override a more specific deny

Implementing these controls independently inside WASM actions is possible, but it makes common authorization rules harder to inspect, validate, test, and audit consistently.

## Proposed Behavior

Introduce a declarative action-policy model evaluated in the proxy before an action is forwarded upstream.

A policy should have:

- A stable rule identifier
- An optional description
- An effect: `allow` or `deny`
- One or more action selectors
- Optional structured request predicates
- An optional reason code and message
- Optional labels or metadata

## Action Selectors

The initial implementation should support matching on:

- Namespace
- MCP method
- Target type: tool, resource, or prompt
- Exact target name
- Validated glob-style target name
- Registry labels
- Resource URI or URI prefix
- Optional authenticated actor attributes when identity is available

The model should be extensible to inference models and agent-to-agent recipients without changing existing rule semantics.

## Request Predicates

For structured MCP parameters, policies should support bounded predicates using JSON Pointer paths.

The initial predicate operators should be deliberately small and deterministic:

- `exists`
- `equals`
- `not_equals`
- `one_of`
- `not_one_of`

Predicates should use JSON value semantics rather than converting all values to strings. Invalid JSON Pointers, unsupported operators, and incompatible operand types must be rejected during configuration validation.

Regular expressions, arbitrary scripts, and general-purpose policy expressions are outside the initial scope.

## Decision Semantics

Wanaku should evaluate all rules applicable to an action.

The decision algorithm must guarantee:

1. If any matching rule has a `deny` effect, the action is denied.
2. A matching allow rule must never override a matching deny rule.
3. Static allow rules must not bypass evaluators or other downstream governance filters.
4. The result must be independent of configuration ordering.
5. Every decision must identify the matched rule or rules.
6. A denial must return a stable machine-readable reason code and a safe human-readable explanation.

Behavior when no rule matches should be controlled by the policy baseline rather than inferred from rule order. Restrictive defaults and audit-only behavior can be addressed by the related fail-safe policy semantics work.

## Pipeline Integration

Static action-policy evaluation should occur after MCP metadata and namespace resolution but before upstream forwarding.

For `tools/call`, policy evaluation should have access to:

- Namespace
- Tool name
- Tool registry entry and labels
- Tool arguments

Equivalent context should be provided for:

- `resources/read`
- `prompts/get`

The implementation should avoid parsing the same request body independently in multiple filters. Existing MCP metadata or a shared parsed request context should be extended where appropriate.

## Configuration and Management

Policies should be configurable through `wanaku.yaml` and inspectable through the management API.

The management API should expose:

- The effective policy configuration
- Individual rule identifiers
- Validation errors with rule-level context
- The rules that contributed to a decision, where authorized

When versioned evaluator or policy configuration is available through #1868, action-policy updates should participate in the same safe activation and revision lifecycle.

## Audit Integration

Policy decisions should produce structured information suitable for the audit trail proposed in #1869.

Audit data should include:

- Rule identifiers
- Final decision
- Stable reason code
- Namespace
- Operation and normalized target
- Policy revision when available

Tool arguments and other sensitive request values should not be copied into audit events by default.

## Scope

The first implementation should cover:

- `tools/call`
- `resources/read`
- `prompts/get`
- Exact and glob target matching
- Registry-label matching
- Bounded JSON Pointer predicates
- Deterministic deny-overrides-allow evaluation

## Non-Goals

- A general-purpose policy language
- Arbitrary code execution in policy predicates
- HTTP or TLS traffic inspection
- Identity-provider integration
- Automatic data classification
- Automated policy recommendations
- Formal policy verification
- A policy-authoring UI
- Replacing contextual LLM and WASM evaluators

## Acceptance Criteria

- [ ] A documented action-policy schema is available.
- [ ] Every rule has a stable identifier and an `allow` or `deny` effect.
- [ ] Policies can match namespace and MCP method.
- [ ] Policies can match tool, resource, and prompt targets.
- [ ] Exact and validated glob-style target matching are supported.
- [ ] Registry-label matching is supported.
- [ ] Structured parameters can be matched using bounded JSON Pointer predicates.
- [ ] Invalid rule identifiers, selectors, pointers, and predicates are rejected at load time.
- [ ] All applicable rules are evaluated.
- [ ] A matching deny always takes precedence over matching allows.
- [ ] Decision results are independent of rule declaration order.
- [ ] Static allows do not bypass evaluator or downstream governance filters.
- [ ] Denials return stable reason codes and safe explanations.
- [ ] Matched rule identifiers are available for audit integration.
- [ ] Policies can be loaded from `wanaku.yaml`.
- [ ] Effective policies can be inspected through the management API.
- [ ] Unit tests cover each selector and predicate operator.
- [ ] Tests cover overlapping allows and denies in both declaration orders.
- [ ] Integration tests verify allowed and denied tool, resource, and prompt operations.
- [ ] Documentation includes configuration examples and decision semantics.

## Design Considerations

Policy matching is expected to run on every governed action. Parsed selectors and predicates should therefore be validated and prepared at configuration activation time rather than repeatedly compiled during request processing.

Rule evaluation must remain deterministic. Configuration ordering may be retained for display purposes, but it must not change deny precedence.

The policy result should distinguish between:

- Explicit allow
- Explicit deny
- No matching rule
- Invalid or unavailable policy state

This distinction will allow fail-safe behavior and audit-mode operation to be implemented without changing the matching model.

## Open Questions

- Should `tools/list`, `resources/list`, and `prompts/list` hide entries that the caller cannot use?
- Should URI matching support prefixes only or a validated glob syntax?
- Should multiple matching deny reasons be returned, or only a primary reason?
- How should actor attributes be represented before identity enforcement is available?
- Should action policies share the evaluator configuration endpoint or have a dedicated management resource?

### LINKED ISSUE #1900: Integrate action-policy enforcement into the MCP pipeline
## Description

Integrate action-policy enforcement into the MCP pipeline for `tools/call`, `resources/read`, and `prompts/get`.

This issue is part of #1870. Implement the policy capability as a dedicated, transport-neutral `features/action-policy` crate. Use an MCP adapter to build the generic action context.

Run `wanaku_action_policy` after MCP metadata, namespace resolution, and initialization handling. Run it before `wanaku_evaluator`. A denial must return a synthetic JSON-RPC response. An allow must return `Continue` so evaluators and downstream governance filters still run.

Do not filter `tools/list`, `resources/list`, or `prompts/list` in this implementation. Document that discovery does not grant authorization.

## Expected Behavior

- The feature loads a compiled immutable policy snapshot through a pipeline extension.
- Tool, resource, and prompt actions contain the required normalized policy context.
- A static denial prevents evaluator and upstream execution.
- A static allow continues through the evaluator and forwarding filters.
- List operations remain unchanged.
- Integration tests verify filter order and upstream isolation after denial.

### LINKED ISSUE #1901: Extract reusable configuration revision lifecycle infrastructure
## Description

Extract the reusable configuration revision lifecycle needed by action policies.

This issue is part of #1870 and builds on the evaluator revision lifecycle delivered by #1868.

Extract only stable lifecycle mechanics, such as bounded immutable history, activation metadata, checksums, optimistic concurrency, rollback, and persistence contracts. Keep feature-specific validation, compilation, and atomic runtime activation in each feature.

Evaluator and action-policy configurations must use separate revision streams, active revision identifiers, concurrency tokens, and persistence data.

## Expected Behavior

- Existing evaluator behavior and management API contracts do not change.
- Existing evaluator revision tests continue to pass.
- Action policy can use the shared lifecycle without depending on evaluator definitions.
- Policy updates cannot create evaluator revision conflicts.
- Evaluator rollback cannot change the active action-policy revision, and the reverse is also true.

### LINKED ISSUE #1902: Define the action-policy schema and matcher interfaces
## Description

Define the declarative action-policy schema and its compiled matcher interfaces.

This issue is part of #1870. Follow the design decisions recorded in the parent issue.

The schema must cover stable rule identifiers, descriptions, allow and deny effects, action selectors, typed JSON Pointer predicates, safe reason codes, and rule metadata. The matcher design must support exact and literal-prefix URI matching in the first implementation. It must permit later matcher implementations without changing established policy semantics.

Missing JSON Pointer values must not match value-comparison operators. The schema must support `exists: false` to match an omitted value. Missing values and JSON `null` must remain distinct.

Reference #1872 for enforcement posture and no-match behavior. Reference #1875 for future validation and simulation consumers.

## Expected Behavior

- The schema has deterministic serialization and validation rules.
- Rule identifiers are unique and stable.
- Unsupported matcher types, operators, operands, and JSON Pointers fail validation.
- Exact and literal-prefix URI matchers compile during policy activation.
- Matcher interfaces permit future glob, regex, and naming-pattern implementations.
- Unit tests cover all selectors, predicates, missing values, `null`, false, zero, and empty strings.

### LINKED ISSUE #1903: Add action-policy integration tests and documentation
## Description

Add end-to-end tests and documentation for the complete action-policy capability from #1870.

This issue must verify the integrated schema, engine, MCP adapter, filter order, management API, revisions, and persistence. Audit-event testing remains in #1869. Future A2A integration remains in #383.

Use ASD-STE100 Simplified Technical English for procedures, configuration, endpoints, system behavior, and other accuracy-critical technical content.

## Expected Behavior

- Integration tests cover allowed and denied tool calls, resource reads, and prompt retrieval.
- Tests cover exact and prefix URI matching.
- Tests cover every JSON Pointer predicate and the missing-value truth table.
- Tests prove deny precedence in multiple rule declaration orders.
- Tests prove that allow continues to the evaluator.
- Tests prove that deny prevents evaluator and upstream execution.
- Tests cover management activation, conflicts, rollback, invalid revisions, and restart persistence.
- Documentation defines the schema, configuration, pipeline order, decision semantics, and list-operation behavior.

### LINKED ISSUE #1904: Add a fallible shared MCP request-view API
## Description

Add a fallible, reusable MCP request-view API for policy adapters and existing MCP filters.

This issue is part of #1870 and builds on the shared JSON-RPC parameter parsing introduced by #1847.

The API must preserve JSON value types. It must distinguish malformed JSON-RPC input, missing parameters, omitted values, and explicit JSON `null`. Do not duplicate body parsing in the action-policy feature.

## Expected Behavior

- The API returns a typed error for malformed governed requests.
- Tool names, resource URIs, prompt names, and structured arguments remain available without string conversion.
- Missing parameters do not silently become a valid empty request view.
- Existing MCP filters use the shared parsing behavior where applicable.
- Unit tests cover valid, malformed, missing, and type-incompatible inputs.

### LINKED ISSUE #1905: Implement the transport-neutral action-policy evaluation engine
## Description

Implement the transport-neutral action-policy evaluation engine from #1870.

The engine must consume a generic action context instead of MCP request types. It must evaluate all applicable rules and apply deterministic deny-overrides-allow semantics. The result must distinguish explicit allow, explicit deny, no match, and unavailable or invalid policy state.

Return one deterministic, safe denial reason to the caller. Retain all contributing rule identifiers and reason codes in the internal decision result.

Define a trusted actor-context interface for future identity integration. Do not expose actor selectors in this implementation. Reference #503 and #1331 for later identity and tenant authorization work.

## Expected Behavior

- Rule declaration order does not change the decision.
- Any matching deny takes precedence over all matching allows.
- An allow does not bypass later governance stages.
- The primary denial reason uses a documented deterministic selection rule.
- The generic action context can support MCP and future A2A adapters.
- Unit tests cover overlapping rules in multiple declaration orders.

### LINKED ISSUE #1906: Add the action-policy management API and persistence
## Description

Add the dedicated action-policy management resource, revision operations, and persistence required by #1870.

Provide action-policy configuration through `wanaku.yaml`. Provide dedicated management routes for effective policy inspection, revision history, activation, rollback, and rule-level validation failures. Do not add policy documents to the evaluator resource or revision stream.

Reference #1872 for posture and readiness behavior. Reference #1875 for future side-effect-free candidate validation and simulation. Reference #1876 for future connector-profile policy composition.

## Expected Behavior

- Action policies have a dedicated `/api/v1/action-policies` resource family.
- Invalid revisions do not replace the last known good revision.
- Optimistic concurrency detects conflicting updates.
- Revision history and the active revision survive restart when persistence is enabled.
- Effective policies and rule identifiers are inspectable through the management API.
- Validation errors identify the applicable rule without exposing sensitive values.

orpiske:
## Summary

- add an extensible action-policy schema, selector matcher interface, and typed JSON predicates
- add deterministic deny-overrides evaluation and fallible MCP request views
- enforce policies before the evaluator for tools/call, resources/read, and prompts/get
- add independent revision history, persistence, and management API routes
- add a read-only Admin UI for active policy details and revision history
- document the policy model, current fail-closed posture, deferred integrations, and API

## Validation

- cargo build
- cargo test
- yarn build (ui/admin)
- focused Action Policies ESLint and TypeScript checks
- Playwright Action Policies suite (2 tests)
- git diff --check upstream/main..HEAD

Closes #1870
Closes #1900
Closes #1901
Closes #1902
Closes #1903
Closes #1904
Closes #1905
Closes #1906

Related to #1869, #383, #1872, #1875, #503, and #1331.

Generated by Codex via OSS Helper

### DIFF
--- a/docs/action-policies.md
+++ b/docs/action-policies.md
@@ -1,0 +1,215 @@
+# Action Policies
+
+Action policies give Wanaku a deterministic authorization layer for MCP actions. They complement the evaluator. Use action policies for rules that depend on known names, labels, URIs, and structured request values. Use evaluators for contextual or semantic decisions.
+
+## Configure a policy
+
+Add one `action_policy` object to `wanaku.yaml`:
+
+```yaml
+action_policy:
+  rules:
+    - id: deny-production-delete
+      description: Prevent destructive production calls
+      effect: deny
+      selectors:
+        namespace: production
+        operation: tools/call
+        target_type: tool
+        target_name:
+          matcher: glob
+          value: "delete-*"
+        labels:
+          risk: high
+      predicates:
+        - operator: equals
+          pointer: /arguments/force
+          value: true
+      reason_code: destructive_action_denied
+      message: This action is not permitted.
+      metadata:
+        owner: platform-security
+
+    - id: allow-public-reports
+      effect: allow
+      selectors:
+        operation: resources/read
+        target_type: resource
+        uri:
+          matcher: prefix
+          value: "https://reports.example/public/"
+```
+
+Wanaku validates and compiles the complete policy before activation. Wanaku rejects the revision if a rule is invalid. An invalid revision does not replace the last active revision.
+
+Each rule uses these fields:
+
+| Field | Required | Value |
+| --- | --- | --- |
+| `id` | Yes | A stable identifier. Start with a letter or digit. Then use letters, digits, `_`, `-`, `.`, `/`, or `:`. |
+| `description` | No | Text for operators. |
+| `effect` | Yes | `allow` or `deny`. |
+| `selectors` | Yes | One or more action selectors. |
+| `predicates` | No | A list of typed JSON Pointer predicates. |
+| `reason_code` | No | A stable identifier for a denied action. |
+| `message` | No | A safe message that Wanaku can return to the caller. |
+| `metadata` | No | Operator metadata. Metadata keys use the identifier syntax. Values are JSON values. |
+
+Unknown fields cause validation to fail. Rule IDs must be unique in one policy.
+
+## Select actions
+
+A rule can use these selectors:
+
+| Selector | Value |
+| --- | --- |
+| `namespace` | The Wanaku namespace. |
+| `operation` | `tools/call`, `resources/read`, or `prompts/get`. |
+| `target_type` | `tool`, `resource`, or `prompt`. |
+| `target_name` | An `exact` or `glob` match expression. |
+| `labels` | Required registry label key-value pairs. |
+| `uri` | An `exact` or `prefix` match expression for a resource URI. |
+
+All selectors and all predicates in one rule use AND semantics. The rule matches only when every configured condition matches.
+
+For `labels`, the configured labels must be a subset of the registry entry labels. Extra labels on the registry entry do not prevent a match. Prompt entries do not currently supply registry labels to the policy adapter.
+
+### Match target names
+
+Use `exact` to compare the complete target name:
+
+```yaml
+target_name: { matcher: exact, value: delete-account }
+```
+
+Use `glob` for target names. `*` matches zero or more characters. `?` matches one character. A backslash escapes the next character. For example, `report-\*` matches the literal name `report-*`. A trailing backslash is invalid.
+
+URI selectors do not support glob matching. They support `exact` and literal `prefix` matching only.
+
+### Match resource URIs
+
+Wanaku compares the exact URI string from `resources/read`. Wanaku does not decode, resolve, or normalize the URI. Case, escaping, separators, and trailing slashes remain significant.
+
+```yaml
+uri:
+  matcher: prefix
+  value: "s3://company-private/"
+```
+
+A URI selector applies only to a resource read. Tool registry transport URIs are not policy resource URIs.
+
+## Match structured request values
+
+Each predicate reads the MCP `params` object with an RFC 6901 JSON Pointer. JSON types remain distinct. The string `"1"` does not equal the number `1`.
+
+| Operator | Operand | Match condition |
+| --- | --- | --- |
+| `exists` | `value: true` or `false` | The pointer is present or absent. |
+| `equals` | `value` | The present value equals the operand. |
+| `not_equals` | `value` | The present value does not equal the operand. |
+| `one_of` | Non-empty `values` | The present value equals one list value. |
+| `not_one_of` | Non-empty `values` | The present value equals no list value. |
+
+Example:
+
+```yaml
+predicates:
+  - operator: exists
+    pointer: /arguments/approved
+    value: true
+  - operator: one_of
+    pointer: /arguments/environment
+    values: [production, staging]
+```
+
+Missing is different from JSON `null`:
+
+| Input at the pointer | `exists: true` | `exists: false` | `equals: null` | `not_equals: null` | `one_of: [null]` | `not_one_of: [null]` |
+| --- | ---: | ---: | ---: | ---: | ---: | ---: |
+| Missing | No | Yes | No | No | No | No |
+| `null` | Yes | No | Yes | No | Yes | No |
+| `false` | Yes | No | No | Yes | No | Yes |
+| `0` | Yes | No | No | Yes | No | Yes |
+| `""` | Yes | No | No | Yes | No | Yes |
+
+A missing value never matches a comparison operator. Use `exists: false` to match an omitted value.
+
+## Decision behavior
+
+Wanaku evaluates all rules. The declaration order does not change the result.
+
+1. If any matching rule has `effect: deny`, Wanaku denies the action.
+2. If no deny matches and an allow matches, Wanaku records an explicit allow and continues the pipeline.
+3. If no rule matches, Wanaku continues the pipeline. The baseline behavior belongs to [#1872](https://github.com/wanaku-ai/wanaku/issues/1872).
+
+A static allow does not skip the evaluator or another downstream filter. A static deny returns a JSON-RPC error before the evaluator and upstream action handler run.
+
+The caller receives one deterministic safe denial reason. Wanaku selects the matching deny rule with the lexicographically lowest rule ID. A configured `reason_code` and `message` take precedence. Otherwise, Wanaku uses `action_policy_denied` and `The requested action is not allowed.` The decision model retains all matching rule IDs and configured reason codes for authorized internal consumers.
+
+An unconfigured policy continues governed requests. If no active last-known-good policy exists, an invalid configured policy rejects governed requests with `action_policy_invalid`. If an active policy exists, an invalid update or startup policy does not replace it. Wanaku continues to enforce the active last-known-good policy. These are temporary failure semantics. [#1872](https://github.com/wanaku-ai/wanaku/issues/1872) owns the final baseline and failure modes.
+
+Wanaku governs `tools/call`, `resources/read`, and `prompts/get`. Wanaku does not filter `tools/list`, `resources/list`, or `prompts/list`. Discovery does not grant authorization.
+
+## Pipeline position
+
+Place `wanaku_action_policy` after MCP metadata, namespace resolution, and `wanaku_mcp_init`. Place it before `wanaku_evaluator`:
+
+```yaml
+- filter: mcp
+- filter: wanaku_namespace
+- filter: wanaku_mcp_init
+- filter: wanaku_action_policy
+- filter: wanaku_evaluator
+```
+
+The default pipeline already uses this order.
+
+## Manage policy revisions
+
+The management API uses a dedicated policy resource:
+
+| Method | Path | Result |
+| --- | --- | --- |
+| `GET` | `/api/v1/action-policies` | Get the effective policy and active revision metadata. |
+| `PUT` | `/api/v1/action-policies` | Validate and activate a policy. |
+| `GET` | `/api/v1/action-policies/revisions` | List revision metadata, newest first. |
+| `GET` | `/api/v1/action-policies/revisions/active` | Get the active revision and policy. |
+| `GET` | `/api/v1/action-policies/revisions/{id}` | Get one revision and policy. |
+| `POST` | `/api/v1/action-policies/revisions/{id}/activate` | Validate the selected policy and create a new active revision. |
+
+Activate a policy:
+
+```bash
+curl -X PUT http://localhost:8080/api/v1/action-policies \
+  -H 'Content-Type: application/json' \
+  -d '{
+    "policy": {
+      "rules": [{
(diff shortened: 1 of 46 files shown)


docs/action-policies.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Match target names
+
+Use `exact` to compare the complete target name:
+
+```yaml
+target_name: { matcher: exact, value: delete-account }
+```
+
+Use `glob` for target names. `*` matches zero or more characters. `?` matches one character. A backslash escapes the next character. For example, `report-\*` matches the literal name `report-*`. A trailing backslash is invalid.
```
orpiske, line 85: Need to define what exactly is a target name. Is it the tool name? An argument value? etc

orpiske (commented):
## Rules-and-conventions review — PR #1907 (action policies)

Thank you for this substantial and well-organized contribution. The change is cohesive: a transport-neutral policy engine, a fallible `McpRequestView`, deny-overrides evaluation, an independent revision stream extracted into a shared `wanaku_types::revision` module, management routes, and a read-only Admin UI — all backed by extensive unit, integration, and E2E tests. It follows the project conventions closely (route enum + resolver + handlers split, `body_filter_boilerplate!`, `FilterAction::Reject` for synthetic responses, context structs, no `unwrap`/`expect`/`panic` outside tests, `Containerfile` updated for the new crate, docs added and cross-linked).

This is a rules-and-conventions review. It does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud/clippy).

### Verified as correct (spot-checked)
- **Deny-overrides determinism**: matches are sorted by rule ID before selecting the primary deny, so the caller-safe reason is the lexicographically lowest deny rule ID regardless of declaration order (matches the docs). Covered by `deny_overrides_allow_and_is_independent_of_declaration_order`.
- **Fail-closed posture**: missing `ActionPolicyState`/`InMemoryRegistry` extensions, poisoned locks, and `PolicyInvalid` all return a JSON-RPC error; `Unconfigured` continues. Consistent with the documented temporary failure semantics.
- **Evaluator checksum stability**: the extracted `wanaku_types::revision::checksum(&defs)` serializes `&&[EvaluatorDef]`, which produces JSON identical to the previous `serde_json::to_string(defs)`, so existing persisted evaluator revisions keep the same checksums (no history churn).
- **Response envelope**: `json_ok`/`json_err` wrap in `{data, error}`, `customFetch` unwraps `.data`, and the UI hook types line up — no double-wrapping.
- **JSON-Pointer / glob validation**: trailing `~`, non-`0/1` escapes, empty matcher values, glob-on-URI, and prefix-on-target-name are all rejected; the glob DP correctly handles `*`, `?`, and backslash escapes.

### Non-blocking observation
- **Rejected startup revisions accumulate across restarts** (see inline comment on `state.rs`). When `wanaku.yaml` holds a schema-valid but semantically-invalid policy and there is no last-known-good active revision, each restart records another `rejected` revision. The 50-entry bound prevents unbounded growth, but the history fills with duplicate rejections. Consider skipping the re-record when the newest revision is already a rejection with the same checksum.

### Minor note (optional)
- In `filter.rs::evaluate_snapshot`, `mcp.method` is re-read with `unwrap_or_default()` after `handle_body` already validated it is a governed method. Harmless redundancy; could pass the validated method through instead.

Overall this looks ready to merge once the startup-rejection edge case is considered (it is not blocking).

features/action-policy/src/state.rs
```diff
@@ -0,0 +1,362 @@
...
+            Ok(startup) => startup,
+            Err(error) => {
+                tracing::error!(error = %error, "action policy configuration is invalid");
+                self.reinstall_active_or_invalid();
+                return;
+            }
+        };
+        match startup {
+            Some(policy) if self.matches_active(&policy) => self.reinstall_active_or_invalid(),
+            Some(policy) => self.activate_startup(policy),
```
orpiske, line 146: Non-blocking: when the startup policy is schema-valid but fails compilation and no last-known-good active revision exists, `active_revision()` is `None`, so `matches_active` returns `false` and `activate_startup` runs on every restart. Each restart records a fresh `rejected` revision, gradually filling the bounded (50) history with duplicates. Consider short-circuiting when the most recent revision is already a rejection with the same policy checksum, so restarts do not keep appending identical rejections.

PR #1960, opened 2026-09-07
Fix #1872: apply posture to action policies

### LINKED ISSUE #1872: Add explicit enforcement modes and fail-safe policy baselines

orpiske:
## Summary

- load and validate governance posture configuration during startup
- inject one immutable governance configuration into each request pipeline
- apply enforce, audit, disabled, no-match, and failure behavior to action policies
- use fail-safe defaults when no governance section is present
- update the governance, action-policy, and configuration documentation
- add focused posture and startup configuration tests

## Scope

This continuation applies posture to deterministic action policies. Evaluator integration, audit side-effect suppression, readiness APIs, audit events, and posture metrics remain for later reviewable slices.

## Validation

- `cargo +1.96.0 test -p wanaku-types governance`
- `cargo +1.96.0 test -p wanaku-feature-action-policy`
- `cargo +1.96.0 test -p wanaku-server governance_config`
- `git diff --check`
- parsed `wanaku.yaml` successfully

The focused action-policy Clippy run reaches pre-existing unfulfilled lint expectations in `wanaku-infra/src/mcp_client.rs` and `wanaku-infra/src/llm.rs`.

Generated by Codex via OSS Helper

## Summary by Sourcery

Apply validated governance posture configuration to action-policy authorization behavior.

New Features:
- Apply governance posture modes and namespace-specific overrides to deterministic action-policy decisions.
- Load, validate, and propagate immutable governance configuration through request pipelines, with fail-safe defaults when omitted.

Bug Fixes:
- Honor configured no-match and policy-failure behavior instead of treating unmatched or unavailable action policies as implicitly allowed or generically denied.

Enhancements:
- Add audit and disabled handling for action-policy evaluation while preserving evaluator integration for a later change.
- Clarify governance, action-policy, and configuration documentation and provide a sample governance configuration.

Documentation:
- Document startup validation, fail-safe defaults, namespace overrides, and action-policy posture behavior.

Tests:
- Add coverage for governance configuration defaults, namespace overrides, invalid disabled scopes, posture behavior, and audit-mode policy handling.

### DIFF
--- a/features/action-policy/src/filter.rs
+++ b/features/action-policy/src/filter.rs
@@ -4,6 +4,9 @@ use bytes::Bytes;
 use praxis_filter::{FilterAction, FilterError, HttpFilterContext};
 use wanaku_filters::json_rpc::{McpRequestView, RequestViewError};
 use wanaku_infra::registry::InMemoryRegistry;
+use wanaku_types::governance::{
+    EnforcementMode, FailureBehavior, GovernanceConfig, GovernancePosture, NoMatchBehavior,
+};
 use wanaku_types::registry::{DEFAULT_NAMESPACE, ToolRegistry};
 
 use crate::{
@@ -16,6 +19,8 @@ const INVALID_ACTION_REASON_CODE: &str = "invalid_action_request";
 const INVALID_ACTION_MESSAGE: &str = "The action request is invalid.";
 const INVALID_POLICY_REASON_CODE: &str = "action_policy_invalid";
 const INVALID_POLICY_MESSAGE: &str = "The action policy is unavailable.";
+const NO_MATCH_REASON_CODE: &str = "governance_no_match";
+const NO_MATCH_MESSAGE: &str = "No governance policy permits this action.";
 
 wanaku_filters::body_filter_boilerplate!(ActionPolicyFilter, "wanaku_action_policy");
 
@@ -35,15 +40,34 @@ impl ActionPolicyFilter {
         let id = wanaku_filters::response::json_rpc_id_from_metadata(
             ctx.get_metadata(wanaku_filters::MCP_ID_KEY),
         );
+        let namespace = ctx
+            .get_metadata(wanaku_types::NAMESPACE_METADATA_KEY)
+            .unwrap_or(DEFAULT_NAMESPACE);
+        let posture = ctx
+            .extensions
+            .get::<GovernanceConfig>()
+            .cloned()
+            .unwrap_or_default()
+            .resolve(namespace);
+        if posture.mode == EnforcementMode::Disabled {
+            tracing::warn!(
+                namespace,
+                reason = posture.disabled_reason.as_deref().unwrap_or("unspecified"),
+                "governance is disabled for namespace"
+            );
+            return Ok(FilterAction::Continue);
+        }
         let Some(state) = ctx.extensions.get::<ActionPolicyState>() else {
-            return Ok(policy_error(
-                &id,
-                INVALID_POLICY_REASON_CODE,
-                INVALID_POLICY_MESSAGE,
-            ));
+            return Ok(apply_failure(&posture, &id));
         };
         let snapshot = state.snapshot();
-        Ok(evaluate_snapshot(ctx, body.as_ref(), snapshot, &id))
+        Ok(evaluate_snapshot(
+            ctx,
+            body.as_ref(),
+            snapshot,
+            &posture,
+            &id,
+        ))
     }
 }
 
@@ -51,17 +75,16 @@ fn evaluate_snapshot(
     ctx: &HttpFilterContext<'_>,
     body: Option<&Bytes>,
     snapshot: PolicySnapshot,
+    posture: &GovernancePosture,
     id: &serde_json::Value,
 ) -> FilterAction {
     let policy = match snapshot {
         PolicySnapshot::Valid(policy) => policy,
-        PolicySnapshot::Unconfigured => return FilterAction::Continue,
-        PolicySnapshot::Invalid => {
-            return policy_error(id, INVALID_POLICY_REASON_CODE, INVALID_POLICY_MESSAGE);
-        }
+        PolicySnapshot::Unconfigured => return apply_no_match(posture, id),
+        PolicySnapshot::Invalid => return apply_failure(posture, id),
     };
     let Some(registry) = ctx.extensions.get::<InMemoryRegistry>() else {
-        return policy_error(id, INVALID_POLICY_REASON_CODE, INVALID_POLICY_MESSAGE);
+        return apply_failure(posture, id);
     };
     evaluate_request(RequestEvaluation {
         method: ctx
@@ -72,6 +95,7 @@ fn evaluate_snapshot(
             .unwrap_or(DEFAULT_NAMESPACE),
         body,
         policy: &policy,
+        posture,
         registry,
         id,
     })
@@ -87,6 +111,7 @@ struct RequestEvaluation<'a> {
     namespace: &'a str,
     body: Option<&'a Bytes>,
     policy: &'a crate::CompiledPolicy,
+    posture: &'a GovernancePosture,
     registry: &'a InMemoryRegistry,
     id: &'a serde_json::Value,
 }
@@ -97,6 +122,7 @@ fn evaluate_request(input: RequestEvaluation<'_>) -> FilterAction {
         namespace,
         body,
         policy,
+        posture,
         registry,
         id,
     } = input;
@@ -107,6 +133,14 @@ fn evaluate_request(input: RequestEvaluation<'_>) -> FilterAction {
         return policy_error(id, INVALID_ACTION_REASON_CODE, INVALID_ACTION_MESSAGE);
     };
     let decision = PolicyEngine::evaluate(PolicyState::Available(policy), &context);
+    if posture.mode == EnforcementMode::Audit {
+        tracing::info!(
+            namespace,
+            decision = ?decision,
+            "action-policy decision recorded in audit mode"
+        );
+        return FilterAction::Continue;
+    }
     match &decision {
         PolicyDecision::ExplicitDeny { .. } => policy_error(
             id,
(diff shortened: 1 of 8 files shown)


features/action-policy/src/filter.rs
```diff
@@ -469,12 +541,77 @@ mod tests {
...
+            FilterAction::Continue
+        ));
+        assert!(matches!(apply_failure(&audit, &id), FilterAction::Continue));
+    }
+
+    #[test]
+    fn audit_mode_does_not_enforce_policy_denial() {
+        let registry = InMemoryRegistry::new();
+        let policy = compile_policy(&serde_json::json!({
+            "id": "deny", "effect": "deny", "selectors": {"operation": "tools/call"}
```
orpiske: I think it should use wanaku_types::TOOLS_CALL

features/action-policy/src/filter.rs
```diff
@@ -469,12 +541,77 @@ mod tests {
...
+        }));
+        let posture = GovernancePosture {
+            mode: EnforcementMode::Audit,
+            ..GovernancePosture::default()
+        };
+        let body = request("tools/call", &serde_json::json!({"name": "unsafe"}));
```
orpiske: Same: wanaku_types::TOOLS_CALL.

features/action-policy/src/filter.rs
```diff
@@ -469,12 +541,77 @@ mod tests {
...
+        let id = serde_json::json!(7);
+
+        assert!(matches!(
+            evaluate_request(RequestEvaluation {
+                method: "tools/call",
```
orpiske: Same: wanaku_types::TOOLS_CALL

types/src/governance.rs
```diff
@@ -1,3 +1,11 @@
+#![cfg_attr(
```
orpiske, line 1: Why this?

wanaku.yaml
```diff
@@ -2,10 +2,20 @@ forwards:
   - name: "echo-mcp"
     address: "http://localhost:9090/mcp"
     namespace: "test-ns"
+
```
orpiske: Unrelated.

orpiske (commented):
Looks very good, only a few minor things to adjust.

features/action-policy/src/filter.rs
orpiske: Updated this selector to use wanaku_types::TOOLS_CALL.

features/action-policy/src/filter.rs
orpiske: Updated the request construction to use wanaku_types::TOOLS_CALL.

features/action-policy/src/filter.rs
orpiske: Updated RequestEvaluation.method to use wanaku_types::TOOLS_CALL.

types/src/governance.rs
orpiske: This allowance is required only when the openapi feature enables utoipa::ToSchema. Rust 1.96 Clippy reports large_stack_frames in the generated schema implementations for GovernancePosture and GovernancePostureOverride. I verified that removing it restores both warnings. I retained the established module-level pattern used by the other OpenAPI schema modules and clarified the reason text.

wanaku.yaml
orpiske: Removed the unrelated blank-line change.

PR #1963, opened 2026-09-07
fix: listing required field  of tools in the public/mcp/sse call

sum-143-it:
required field is not populating with an actual value, coming as []

## Summary by Sourcery

Bug Fixes:
- Correctly populate each tool argument’s required status in public MCP/SSE tool listings.


orpiske (approved):
LGTM, thanks!

PR #1982, opened 2026-09-14
Add a container image build to cargo

### LINKED ISSUE #1981: Add a container image build to cargo
Add cargo tasks to build the container image, also takes into account the different kubernetes clusters, like openshift, minikube, etc.

claudio4j:
Fix https://github.com/wanaku-ai/wanaku/issues/1981

## Summary by Sourcery

Provide first-class Cargo-driven container image builds for local, Minikube, and OpenShift workflows, including headless and multi-architecture variants.

New Features:
- Add Cargo and Make commands for building full or headless Wanaku container images across Docker, Minikube, and OpenShift environments.
- Support native multi-architecture image builds and manifest assembly.

Bug Fixes:
- Adapt the container image to run under OpenShift restricted security policies.

Enhancements:
- Add an xtask-based container build workflow with environment-specific registry handling and local build validation.

Build:
- Include the xtask package and container build configuration in the workspace and Containerfile.

Deployment:
- Configure the authentication compose service with the standard Wanaku server image tag.

Documentation:
- Document container image build commands, target environments, tagging, multi-architecture workflows, and prerequisites.


.github/workflows/ui-e2e-tests.yml
```diff
@@ -65,7 +65,7 @@ jobs:
 
     - name: Start Server
       run: |
-        cargo run &
+        cargo run --bin wanaku-server &
```
orpiske: I'm not entirely sure about this one ... feels like out of scope, no?

orpiske (changes requested):
Thanks ... I think you are on the right track. I like the way you propose doing it via `cargo` and how we can hide some of the complexities of the image building with it. 

However, there's one problem to solve: this is not multi-plaform. You need to adjust the code so that the build can happen per platform ([like we do today](https://github.com/wanaku-ai/wanaku/blob/main/.github/workflows/main-build.yml#L68-L90)) and a [manifest](https://github.com/wanaku-ai/wanaku/blob/main/.github/workflows/main-build.yml#L68-L90) [that contains all these platforms (`x86_64` and `aarm64`) is created](https://github.com/wanaku-ai/wanaku/blob/main/build-manifests.sh).

Without the manifest, then running `podman run wanaku-img` fails on ARM machines such as modern Macs based on Apple Silicon.

orpiske:
This is what [we have today:](https://quay.io/repository/wanaku/wanaku-server?tab=tags) 

Screenshot 2026-09-14 at 19 10 15

```
wanaku/wanaku-server:latest = {
    wanaku/wanaku-server:main-x86_64,
    wanaku/wanaku-server:main-aarm64
}
```

.github/workflows/ui-e2e-tests.yml
claudio4j: With the addition of the xtask, "cargo run" requires to set it, otherwise it fails
```
error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
available binaries: wanaku-server, xtask
```

claudio4j: I had set the default-run, so "cargo run" just works as usual.

claudio4j:
I will work to add the multiarch parameter.

claudio4j:
multiarch container image builds are introducedd
```
# Machine 1 (x86_64):
cargo build-image --multiarch --tag quay.io/wanaku/wanaku-server:latest
# → builds + pushes quay.io/wanaku/wanaku-server:latest-x86_64

# Machine 2 (aarch64):
cargo build-image --multiarch --tag quay.io/wanaku/wanaku-server:latest
# → builds + pushes quay.io/wanaku/wanaku-server:latest-aarch64

# Either machine — after both are in the registry:
cargo build-image --multiarch --manifest --tag quay.io/wanaku/wanaku-server:latest
# → podman manifest create latest  latest-aarch64 latest-x86_64
# → podman manifest push --all latest
```

orpiske (commented):
This LGTM ... a few more minor things and we can merge:

1. Can you also adjust the CI main build so it uses the `cargo` command instead: https://github.com/wanaku-ai/wanaku/blob/main/.github/workflows/main-build.yml ? 
2. Can you also remove the files `images.txt` and `build-manifests.sh` as they become outdated after this patch? 

Thanks

docs/container-build.md
```diff
@@ -0,0 +1,197 @@
...
+### Multi-arch builds
+
+Multi-arch support mirrors the CI native two-pass approach — no QEMU emulation is used.
+Each machine builds only its own native architecture.
+
+**Step 1 — build and push the arch-specific image on each machine:**
+
+```bash
+# On an x86_64 machine:
+cargo build-image --multiarch --tag quay.io/wanaku/wanaku-server:latest
```
orpiske: Shouldn't this command be `cargo build-image --multiarch --tag quay.io/wanaku/wanaku-server:latest-x86_64` instead of `:latest` ? Also, do we really need the `--multiarch` ? 

I think the build needs to continue to be in 2 steps: 1 to build the host-specific container image, and another to build the manifest.

docs/container-build.md
```diff
@@ -0,0 +1,197 @@
...
+# pushes: quay.io/wanaku/wanaku-server:latest-x86_64
+
+# On an aarch64 machine:
+cargo build-image --multiarch --tag quay.io/wanaku/wanaku-server:latest
```
orpiske: Same thing here: shouldn't it be `:latest-aarch64` ?

docs/container-build.md
```diff
@@ -0,0 +1,197 @@
...
+# pushes: quay.io/wanaku/wanaku-server:latest-aarch64
+```
+
+**Step 2 — assemble the manifest list** (run once, on either machine, after both arch
+images are in the registry):
+
+```bash
+cargo build-image --multiarch --manifest --tag quay.io/wanaku/wanaku-server:latest
```
orpiske: I think this could be a different command: `cargo assemble-manifest`

orpiske (changes requested):
@claudio4j I did one more pass on this and I think I have a few more suggestions ... I think this needs to happen two steps, wich the first one for building the host-specific container and the second to build the manifest. 

I think this would also simplify packing images for other archs like ppc64le and s390x.

docs/container-build.md
claudio4j: the tag already contains the arch suffix, this is the image after build (log output)
```
 => => naming to quay.io/wanaku/wanaku-server:latest-x86_64 
```

```
docker images|grep wanaku
quay.io/wanaku/wanaku-server:latest-x86_64   c0ab646f3dcb        214MB
```

Probably the idea is to remove the multiarch parameter and just build the container image at the host specific, like "cargo build-image" on my x86_64 will build `quay.io/wanaku/wanaku-server:latest-x86_64` and on an mac machine, the build will be  `quay.io/wanaku/wanaku-server:latest-arm64`

Following your suggestion from other comment to have a specific command to build and push only the manifest.

claudio4j: they are ready for review. thanks.

orpiske (approved):
Thanks!

PR #2010, opened 2026-09-20
docs: add a real hosted MCP server as the remote forward example

mouse-value-add:
The getting-started guide only shows forwards against a local `echo-mcp`-style server. Since Wanaku's tool discovery is mostly about connecting to upstream MCP servers, it seemed worth showing one real, hosted endpoint in the docs so new users can see the flow end to end without standing anything up first.

I used You.com's hosted MCP endpoint for the example:

- **Keyless profile** (`https://api.you.com/mcp?profile=free`) — no API key, so `wanaku forwards add` is a one-liner and discovery immediately registers the upstream search tools (e.g., `you-search`) into the local catalog.
- **Authenticated endpoint** (`https://api.you.com/mcp`) — instead of inventing new credential plumbing, the example points at the existing [header-forwarding allowlist](docs/configuration.md#header-forwarding) (`WANAKU_FORWARD_HEADERS=Authorization`), which already covers the gateway-mediated token case.

Docs-only change; no code paths, defaults, or behavior touched. I verified the keyless endpoint with a live MCP handshake (initialize + tools/list) over streamable HTTP, which is the transport the client in `infra/src/mcp_client.rs` uses.

Happy to adjust the wording or drop the authenticated-endpoint paragraph if you'd rather keep the guide to a single example.

## Summary by Sourcery

Documentation:
- Add a hosted You.com MCP server example to the getting-started guide, covering both keyless discovery and authenticated requests through header forwarding.


orpiske:
Spam.

